PR #15, opened 2023-11-16
Move find_all to entity_api

calebbourg:
## Description

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns

### DIFF
--- a/entity_api/src/organization.rs
+++ b/entity_api/src/organization.rs
@@ -1,26 +1,31 @@
 use super::error::Error;
 use entity::organization;
 use organization::{Entity, Model};
-use sea_orm::{entity::prelude::*, ActiveValue, DatabaseConnection};
+use sea_orm::{entity::prelude::*, ActiveValue};
 use serde_json::json;
+use service::AppState;
 
-pub async fn find_all(db: &DatabaseConnection) -> Vec<Model> {
+pub async fn find_all(app_state: &AppState) -> Vec<Model> {
+    let db = app_state.database_connection.as_ref().unwrap();
     Entity::find().all(db).await.unwrap_or(vec![])
 }
 
-pub async fn find_by_id(db: &DatabaseConnection, id: i32) -> Result<Option<Model>, Error> {
+pub async fn find_by_id(app_state: &AppState, id: i32) -> Result<Option<Model>, Error> {
+    let db = app_state.database_connection.as_ref().unwrap();
     Entity::find_by_id(id)
         .one(db)
         .await
         .map_err(|err| err.into())
 }
 
-pub(crate) async fn seed_database(db: &DatabaseConnection) {
+pub(crate) async fn seed_database(app_state: &AppState) {
     let organization_names = [
         "Jim Hodapp Coaching",
         "Caleb Coaching",
         "Enterprise Software",
     ];
+
+    let db = app_state.database_connection.as_ref().unwrap();
 
     for name in organization_names {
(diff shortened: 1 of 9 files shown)


web/src/error.rs
```diff
@@ -4,8 +4,8 @@ use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
 use serde::Serialize;
 
-use entity_api::error::EntityApiErrorType;
-use entity_api::error::Error as EntityApiError;
+use entity_api::error::EntityApiError;
+use entity_api::error::Error as EntityApiErrorSuper;
```
calebbourg, line 8: Let's talk through this when we meet later today. I think you've definitely exposed a part of how this is set up that could use more thought. I'm not sure I love adding the `Super` since to me that implies inheritance or nesting. But I also don't have a better option in mind off-hand.

web/src/controller/organization_controller.rs
```diff
@@ -22,7 +22,9 @@ impl OrganizationController {
     /// --request GET \
     /// http://localhost:4000/organizations
     pub async fn index(State(app_state): State<AppState>) -> impl IntoResponse {
-        let organizations = OrganizationApi::find_all(&app_state).await;
+        let organizations = OrganizationApi::find_all(&app_state)
```
calebbourg, line 25: I think I prefer keeping the `unwrap_or_default`  detail inside of the entity_api context since to me it's an implementation detail and will always be the behavior we're expecting with a `find_all`. Open to your thoughts on this.

entity_api/src/organization.rs
```diff
@@ -5,17 +5,14 @@ use sea_orm::{entity::prelude::*, ActiveValue};
...
+    Ok(Entity::find().all(db).await?)
 }
 
 pub async fn find_by_id(app_state: &AppState, id: i32) -> Result<Option<Model>, Error> {
     let db = app_state.database_connection.as_ref().unwrap();
-    Entity::find_by_id(id)
-        .one(db)
-        .await
-        .map_err(|err| err.into())
+    Ok(Entity::find_by_id(id).one(db).await?)
```
calebbourg, line 15: I like this a lot. I'm curious, though, if we're now only transforming the error up one level in the actual controller? I think we'd need to write a test to determine this.

jhodapp: I think you figured this out better in your other branch. Wrapping in `Ok()` is a valid solution because the `Result<>` is unwrapped from await(), but we need it to be in a `Result<>` for both the `Ok` or `Err` cases.

But having a naked `?` operator is always going to be my favorite solution syntactically.

web/src/controller/organization_controller.rs
jhodapp: That's ok, this was only recognizing that `unwrap_or(vec![])` vs `unwrap_or_default()` should return the same thing. But if you think moving any unwrap into the EntityAPI for Organization is even stronger, I'm good with that. Let's take a look at this further together.

web/src/error.rs
jhodapp: Indeed, I just borrowed the `Super` nomenclature from Rust modules which I think makes this fair. My approach also localizes the name distinction to this module only instead of EntityApiErrorType having "Type" always. From my understanding of Rust convention in the community, adding "Type" to the end of a type is not a common practice. It is in C/C++ land, but not Rust.

calebbourg:
Closing in favor of #21

PR #18, opened 2023-11-18
Add two new CLI flags to set the host side TCP interface and port

jhodapp:
## Description
This PR adds two new CLI flags to set the host side TCP interface and port to listen on for incoming client connections. Sets the default port to 4000.

### Changes
* Adds CLI flag `-i` as `String` to set the host side TCP interface address
* Adds CLI flag `-p` as `u16` to set the host side TCP port

### Testing Strategy
Run the following:
--> `cargo run -- -t 2 -p 3000 -i 127.0.0.1`

and observe that the server outputs an INFO log line showing:
--> `[INFO] Server starting... listening for connections on http://127.0.0.1:3000`

### Concerns
None

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -14,16 +14,16 @@ mod router;
 
 pub async fn init_server(app_state: AppState) -> Result<()> {
     // These will probably come from app_state.config (command line)
-    let host = "127.0.0.1";
-    let port = 3000;
+    let host = app_state.config.interface.as_ref().unwrap();
+    let port = app_state.config.port;
     let server_url = format!("{host}:{port}");
 
-    let addr = SocketAddr::from_str(&server_url).unwrap();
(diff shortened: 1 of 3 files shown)


web/src/lib.rs
```diff
@@ -14,16 +14,16 @@ mod router;
 
 pub async fn init_server(app_state: AppState) -> Result<()> {
     // These will probably come from app_state.config (command line)
-    let host = "127.0.0.1";
-    let port = 3000;
+    let host = app_state.config.interface.as_ref().unwrap();
```
calebbourg, line 17: Curious why we need the as_ref here?

calebbourg (approved):

web/src/lib.rs
jhodapp: The unwrap() consumes the value otherwise and rustc complains about the ownership move farther down in the code on line 27.

PR #29, opened 2024-01-15
axum-login credential-checking and user session creation

jhodapp:
## Description
Adds axum-login based credential-checking and user session creation/checking/deletion available at two specific new endpoints:

`/login`
`/logout`

### Changes
* Adds a password field to the `User` entity. Make sure to rerun DB migrations.
* Protects all existing entity-api endpoints behind needing to log in and create a session
* Allows 2 DB-seeded users to log in and can maintain the 2 distinct sessions
* The entity-api endpoints do work, e.g. organizations/*, but only if you pass the session id with every request
* The user can log out thus destroying the session
* Uses PostgresStore to store created user sessions, currently stores them under `tower_sessions/sessions`

### Testing Strategy
* Log in: `curl -v --header "Content-Type: application/x-www-form-urlencoded" --data "username=james.hodapp@gmail.com&password=password1&next=organizations" http://localhost:4000/login`
* Notice the cookie that is create and mentioned in the reply which includes a session ID. Copy and replace <session_id> for all other requests.
* Test an organization endpoint: `curl -v --header "Cookie: id=<session_id>" --request GET http://localhost:4000/organizations`
* This should work as expected
* Log out: `curl -v --header "Cookie: id=<session_id>" --request GET http://localhost:4000/logout`

### Concerns
- The existing endpoints tests are broken, they need to be updated to log in and create/use a session


web/src/controller/user_session_controller.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// http://localhost:4000/login
+    pub async fn get_login(
+        Query(NextUrl { next }): Query<NextUrl>,
+    ) -> Result<impl IntoResponse, Error> {
+        debug!(
+            "UserSessionController::get_login(), next: {:?}",
+            next.unwrap_or_default()
+        );
+
+        // TODO: try and respond with an HTML template like the example until we understand
```
calebbourg, line 84: I don't _believe_ we will need this as it will be up to the FE to render a form for logging a user in. 

The form will then take the user's credentials and make a call to `/login` above.

jhodapp: Yeah you're absolutely right, this comes from the examples I was referencing and is clearly meant for server side rendering of pages and not a React-based frontend.

jhodapp: I removed this.

entity_api/src/user.rs
```diff
@@ -0,0 +1,109 @@
+use async_trait::async_trait;
+use axum_login::{AuthnBackend, UserId};
+use entity::user::{self, Column, Model};
+//use super::error::{EntityApiErrorCode, Error};
```
jhodapp, line 4:
```suggestion
```

entity_api/src/user.rs
```diff
@@ -0,0 +1,109 @@
...
+        Self {
+            db: Arc::new(db.clone()),
+        }
+    }
+}
+
+#[async_trait]
+impl AuthnBackend for Backend {
+    // TODO: I think we need to wrap entity::user::Model so that
+    // the DeriveEntityModel doesn't "erase" the AuthUser impl
```
jhodapp, line 35:
```suggestion
```

jhodapp:
@calebbourg expressed that this looks like a solid enough base to not just use as a PoC but to actually land this code.

So the main thing that needs to be done before landing is to update the failing tests so that they're able to work with the login/session system. And perhaps add a new blackbox test or two for the new login/user session endpoints.

calebbourg:
Approving despite CI failure. Issue tracked in https://github.com/Jim-Hodapp-Coaching/refactor-platform-rs/issues/32

calebbourg (approved):

PR #36, opened 2024-02-26
Fix all router endpoint integration tests

### LINKED ISSUE #32: Fix ability to run tests
### Contact Details

_No response_

### What happened?

We are unable to successfully compile the app in test runs. 

### What is the version of refactor-platform-rs that you're seeing the problem on?

0.3

### Any related code to produce this issue and relevant log output.

_No response_

jhodapp:
## Description
This PR fixes all of the broken tests in router.rs that broke when we introduced AuthSession from axum-login and the Organization endpoints being protected behind a valid login session.

### Changes
* Update all tests to log in and establish a valid AuthSession

### Testing Strategy
1. Run `cargo test` and ensure all tests pass

### Concerns
None

### DIFF
--- a/web/src/router.rs
+++ b/web/src/router.rs
@@ -55,6 +55,7 @@ pub fn static_routes() -> Router {
 #[cfg(feature = "mock")]
 mod organization_endpoints_tests {
     use super::*;
+    use anyhow::Ok;
     use axum_login::{
         tower_sessions::{Expiry, MemoryStore, SessionManagerLayer},
         AuthManagerLayerBuilder,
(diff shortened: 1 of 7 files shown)


web/src/router.rs
```diff
@@ -103,39 +125,95 @@ mod organization_endpoints_tests {
...
+                })
+                .to_string()
+            );
+
+            Ok(())
+        }
+
+        /// Creates a test user::Model entity instance that can be used by tests to
+        /// log in to the /login endpoint and create a valid AuthSession.
+        pub fn get_user() -> anyhow::Result<user::Model> {
```
jhodapp, line 161: @calebbourg what do you think of this function name `get_user()`? Do you think it's confusing such that you'd expect some more dynamic getter of a property instead of generating a static instance of a `user::Model`?

Would `create_user()` or `create_test_user()` be better or do you think it's fine as is and I'm just overthinking it?

calebbourg (approved):

PR #44, opened 2024-03-20
Add a custom Extractor to check for a valid API version in client request header

### LINKED ISSUE #43: Introduce API versioning
**Description:**

We've reached the point where we need to introduce a simple API versioning scheme so that we're not trying to add this in after the codebase (and API) is much further developed.

We've decided to use semantic versioning, so we'll be using 3 octets style. The first version will be set to "0.0.1" and remain there until we've reached a stable version of our API. At that point we'll change it to "1.0.0".

Caleb and I also decided to use the RESTful versioning style of header-based versioning instead of URL path or query string versioning.

**Definition of Done:**
- Can specify the API version to use on the command line and in an environment variable
- All endpoints are exported under the version "0.0.1" such that, if one specified any other version, there would be no exported endpoints

jhodapp:
## Description
Adds a custom Extractor that checks for a valid API version set in a client request header before calling the endpoint's handler. This enables us to call certain route handlers for different versions, and allows us to know the current version in a handler and do different functionality depending on the version number.

### Changes
* Adds a new CheckApiExtractor that looks for the `X-Version` header and compares that to the current version of the API we've set in `AppState.config.api_version`
* Modifies tests to include this new client header

### Testing Strategy
Run `cargo test`, if it passes it's successfully used the `X-Version` header

### Concerns
None

### DIFF
--- a/web/src/custom_extractors.rs
+++ b/web/src/custom_extractors.rs
@@ -28,10 +28,15 @@ where
     async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
         let state = AppState::from_ref(state);
         let version = get_x_version(parts)?;
-        let api_version = HeaderValue::from_str(&state.config.api_version.unwrap_or_default()).ok().unwrap();
+        let api_version = HeaderValue::from_str(&state.config.api_version.unwrap_or_default())
+            .ok()
+            .unwrap_or_else(|| HeaderValue::from_static("0.0.0"));
 
         trace!("API version provided by client: {:?}", version);
-        trace!("API version set in AppState.config.api_version: {:?}", version);
+        trace!(
+            "API version set in AppState.config.api_version: {:?}",
+            version
+        );
 
         Ok(is_current_api_version(version, api_version)?)
     }
@@ -45,7 +50,10 @@ fn get_x_version(parts: &mut Parts) -> Result<HeaderValue, RejectionType> {
     }
 }
 
-fn is_current_api_version(version: HeaderValue, api_version: HeaderValue) -> Result<CheckApiVersion, RejectionType> {
+fn is_current_api_version(
+    version: HeaderValue,
+    api_version: HeaderValue,
+) -> Result<CheckApiVersion, RejectionType> {
     if version == api_version {
         Ok(CheckApiVersion(version))
     } else {
--- a/web/src/custom_extractors.rs
+++ b/web/src/custom_extractors.rs
@@ -6,6 +6,8 @@ use axum::{
 };
 use log::*;
 
+type RejectionType = (StatusCode, &'static str);
+
 pub static X_VERSION: &str = "x-version";
 
 pub struct CheckApiVersion(pub HeaderValue);
@@ -16,7 +18,7 @@ where
     AppState: FromRef<S>,
     S: Send + Sync,
 {
-    type Rejection = (StatusCode, &'static str);
+    type Rejection = RejectionType;
 
     // A custom Extractor that extracts and checks that the API version number
     // provided in the "X-Version" header is equal to the API version specified
@@ -25,18 +27,31 @@ where
     // successfully.
     async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
         let state = AppState::from_ref(state);
+        let version = get_x_version(parts)?;
+        let api_version = HeaderValue::from_str(&state.config.api_version.unwrap_or_default()).ok().unwrap();
 
-        if let Some(version) = parts.headers.get(X_VERSION) {
-            trace!("API version provided by client: {:?}", version);
-            match state.config.api_version == Some(version.to_str().unwrap_or_default().into()) {
-                true => Ok(CheckApiVersion(version.clone())),
-                false => Err((
-                    StatusCode::BAD_REQUEST,
-                    "`X-Version` header is not a valid API version",
-                )),
-            }
-        } else {
-            Err((StatusCode::BAD_REQUEST, "`X-Version` header is missing"))
-        }
+        trace!("API version provided by client: {:?}", version);
+        trace!("API version set in AppState.config.api_version: {:?}", version);
+
+        Ok(is_current_api_version(version, api_version)?)
+    }
+}
+
+fn get_x_version(parts: &mut Parts) -> Result<HeaderValue, RejectionType> {
+    if let Some(version) = parts.headers.get(X_VERSION) {
+        Ok(version.clone())
+    } else {
+        Err((StatusCode::BAD_REQUEST, "`X-Version` header is missing"))
+    }
+}
+
+fn is_current_api_version(version: HeaderValue, api_version: HeaderValue) -> Result<CheckApiVersion, RejectionType> {
+    if version == api_version {
+        Ok(CheckApiVersion(version))
+    } else {
+        Err((
+            StatusCode::BAD_REQUEST,
+            "`X-Version` header is not a valid API version",
+        ))
     }
 }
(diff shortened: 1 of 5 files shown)


web/src/custom_extractors.rs
```diff
@@ -0,0 +1,42 @@
...
+                true => Ok(CheckApiVersion(version.clone())),
+                false => Err((
+                    StatusCode::BAD_REQUEST,
+                    "`X-Version` header is not a valid API version",
+                )),
+            }
+        } else {
+            Err((StatusCode::BAD_REQUEST, "`X-Version` header is missing"))
+        }
+    }
```
calebbourg, line 41: I think you could potentially rewrite this to not have the `match` nested inside of an `if/else`.

A `match` with `true` and `false` as the 2 arms is essentially an `if/else` statement. In other languages with this sort of construct the idiom would be to use the `if/else`.
 So maybe we could move the inner conditional into it's own function and call that inside the affirmative branch of the outer conditional and use the `?` operator to return the success result or `Err`

calebbourg (commented):
This is really sweet! I love the use of a custom extractor here.

What were you thinking in terms of how this will be used in the wild? For example, in the case that we want to support multiple versions simultaneously (let's say `0.1` and `0.2`), we could deploy two instances each with their respective version variable set. 

In the code we would either need to:
- Check the version and have conditional logic to ensure the app behaves like `0.1` vs `0.2`.
  - This would probably require us to maintain some sort of table mapping a given version to it's respective code path.
  - This would allow us to always deploy a single version of the application itself and that version would support all supported API versions
OR
- Maintain a separate branch for each
  - This would allow us to patch and develop separate versions cleanly
  - This would require us to maintain multiple deployments (at least one per supported app version)

My thoughts would be the former as I don't foresee the need to support many versions simultaneously anytime soon but I'm curious on your thoughts and if this is aligned with what your intentions were.

jhodapp:
> This is really sweet! I love the use of a custom extractor here.

Thanks! It took some digging to figure out the best approach, but I think this is the way to go. It makes the handlers calling function a bit like Elixir's function pattern matching in that the extractor(s) must not fail if the handler is to be fully called.

> 
> What were you thinking in terms of how this will be used in the wild? For example, in the case that we want to support multiple versions simultaneously (let's say `0.1` and `0.2`), we could deploy two instances each with their respective version variable set.
> 
> In the code we would either need to:
> 
> * Check the version and have conditional logic to ensure the app behaves like `0.1` vs `0.2`.
>   
>   * This would probably require us to maintain some sort of table mapping a given version to it's respective code path.
>   * This would allow us to always deploy a single version of the application itself and that version would support all supported API versions

I'm thinking of this approach - I'm not a fan of maintaining multiple stable branches of the same codebase as it gets to be a maintenance burden quickly. We can string together handlers using [or](https://docs.rs/axum-extra/latest/axum_extra/handler/trait.HandlerCallWithExtractors.html#method.or) from axum-extra. Using `or()`, if one handler fails because the version extractor fails, it'll try the next one in order. These would be used for major version changes to the API itself.

Since we get access to the version in the way I wrote the extractor, we can also use this version to do different things (version logic) in the same handler if the API didn't change but something about its internal logic changed that needs to be distinct by version. These would used for be minor or patch level changes to a single handler.

>     OR
> * Maintain a separate branch for each
>   
>   * This would allow us to patch and develop separate versions cleanly
>   * This would require us to maintain multiple deployments (at least one per supported app version)
> 
> My thoughts would be the former as I don't foresee the need to support many versions simultaneously anytime soon but I'm curious on your thoughts and if this is aligned with what your intentions were.

Hopefully this all makes sense?

web/src/custom_extractors.rs
jhodapp: Good catch, I originally had nested ifs but you're right, this can be taken further thanks to the versions being wrapped in `Option`. I'll see what I can do.

calebbourg:
@jhodapp Thank you for the explanation!

I agree that maintaining multiple branches sounds complicated and potentially more troublesome than worth.

I've been thinking a bit more about the use of an extractor here and I'm not fully convinced that this is where we want the check to be. A couple main thoughts come to mind:
- Other extractors provided by axum have names that are nouns. Ex. `Json`, `Path`, `State`. This fits with what these extractors are doing which is "provide this value". `Json` evaluates to a JSON value. `Path` evaluates to hte path. I think in our case we would expect to name the custom extractor `ApiVersion` and it's sole purpose would be to take the incoming request and evaluate to whatever the `ApiVersion` is.
- I think it's arguable that the other extractors only succeed given valid `Path`, `State`, etc. information and so likely do some types of checks on their own and we _could_ consider an API version that we no longer support as "invalid data"
- With this pattern we will need to add this extractor to every handler. Although not by any means, the end of the world, this gives me a bit of pause and makes me wonder if there is a better place to perform the check.

I think there are two separate concerns here potentially:
1. What is the API version requested by the client.
2. Is the API version requested by the client supported by our BE

A custom extractor called `ApiVersion` without the specific check could satisfy the first concern.

Would it be possible to move use an `ApiVersion` extractor upstream to check _all_ incoming requests for a supported version and then provide a unified error response from one place in the code? Potentially as a middleware component?

The we could still use  the `ApiVersion` extractor in specific routes/handlers to provide the kind of fine-grained branching logic you described (which I think would definitely provide value).

jhodapp:
> @jhodapp Thank you for the explanation!

@calebbourg You're welcome, thanks for the thought-provoking follow-up reflections!

> 
> I agree that maintaining multiple branches sounds complicated and potentially more troublesome than worth.
> 
> I've been thinking a bit more about the use of an extractor here and I'm not fully convinced that this is where we want the check to be. A couple main thoughts come to mind:
> 
> * Other extractors provided by axum have names that are nouns. Ex. `Json`, `Path`, `State`. This fits with what these extractors are doing which is "provide this value". `Json` evaluates to a JSON value. `Path` evaluates to hte path. I think in our case we would expect to name the custom extractor `ApiVersion` and it's sole purpose would be to take the incoming request and evaluate to whatever the `ApiVersion` is.

Actually not entirely true. The example the Axum docs gives for [writing a custom extractor](https://docs.rs/axum/latest/axum/extract/index.html#defining-custom-extractors) uses the verb "Extract" in the name "ExtractUserAgent".

I did agree the name could be slightly better, so I renamed it to "CompareApiVersion". Let me know what you think of that.

> * I think it's arguable that the other extractors only succeed given valid `Path`, `State`, etc. information and so likely do some types of checks on their own and we _could_ consider an API version that we no longer support as "invalid data"

100% agree.

> * With this pattern we will need to add this extractor to every handler. Although not by any means, the end of the world, this gives me a bit of pause and makes me wonder if there is a better place to perform the check.
> 
> I think there are two separate concerns here potentially:
> 
> 1. What is the API version requested by the client.
> 2. Is the API version requested by the client supported by our BE
> 
> A custom extractor called `ApiVersion` without the specific check could satisfy the first concern.

This would absolutely be possible and my preferred approach if Axum allowed for route endpoints that have the same path but are routed to different handlers depending on criteria (like Extractors). But it doesn't allow this unfortunately (I read a recent long thread on this from one of Axum's founders).

On the point of repeating the specific API check in each handler with the extractor is also already part of the extractor pattern that Axum sets. For example, the `State` extractor is listed in every one of our controller handlers already.

> 
> Would it be possible to move use an `ApiVersion` extractor upstream to check _all_ incoming requests for a supported version and then provide a unified error response from one place in the code? Potentially as a middleware component?

This would somewhat be possible if we used the same first tier handler for all endpoints which would then do some sub-routing to the specific second tier handler (i.e. our controller handlers). I believe this would be engineering against the design pattern of Axum's router.

> 
> The we could still use the `ApiVersion` extractor in specific routes/handlers to provide the kind of fine-grained branching logic you described (which I think would definitely provide value).

If I understand what you're thinking, we'll be able to accomplish this through the use of `[Or]([or](https://docs.rs/axum-extra/latest/axum_extra/handler/trait.HandlerCallWithExtractors.html#method.or))` from axum-extras

calebbourg (approved):
Thank you for walking me through that!

PR #46, opened 2024-05-13
Get organization by external

calebbourg:
## Description
This PR updates the `GET /organizations/{id}` end point to query organizations by `external_id` (`GET /organizations/{external_id}`

### Changes
* Updates organizations read endpoint

### Testing Strategy
I was able to test this using rapidoc
Screen Shot 2024-05-13 at 8 51 18 AM

### DIFF
--- a/entity_api/src/organization.rs
+++ b/entity_api/src/organization.rs
@@ -91,6 +91,18 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>
     Ok(organization)
 }
 
+pub async fn find_by_external_id(
+    db: &DatabaseConnection,
+    external_id: &ExternalId,
+) -> Result<Option<Model>, Error> {
+    let organization = Entity::find()
+        .filter(Column::ExternalId.eq(Uuid::parse_str(external_id).unwrap()))
+        .one(db)
+        .await?;
+
+    Ok(organization)
+}
+
 pub async fn find_by(
     db: &DatabaseConnection,
     params: HashMap<String, String>,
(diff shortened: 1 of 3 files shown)


entity_api/src/organization.rs
```diff
@@ -206,4 +218,23 @@ mod tests {
...
+        let db = MockDatabase::new(DatabaseBackend::Postgres).into_connection();
+
+        let _ = find_by_external_id(&db, &external_id.to_string()).await;
+
+        assert_eq!(
+            db.into_transaction_log(),
+            [Transaction::from_sql_and_values(
+                DatabaseBackend::Postgres,
+                r#"SELECT "organizations"."id", "organizations"."external_id", "organizations"."name", "organizations"."logo", "organizations"."created_at", "organizations"."updated_at" FROM "refactor_platform"."organizations" WHERE "organizations"."external_id" = $1 LIMIT $2"#,
+                [external_id.into(), 1u64.into()]
```
jhodapp, line 234: `1u64` is interesting to compare `external_id` to. Is this right, I would expect it to be a `String`?

calebbourg: Good question! This actually is comparing `1u64` to the `$2` value in `LIMIT $2` which is added by the `.one()` function in

```Rust
    let organization = Entity::find()
        .filter(Column::ExternalId.eq(uuid))
        .one(db)
        .await?;
```
the `into()` call converts the `1u64` into a [BigUnsigned](https://docs.rs/sea-orm/latest/sea_orm/enum.Value.html#variant.BigUnsigned) which is, I believe, how seaORM represents the underlying postgres type

entity_api/src/organization.rs
```diff
@@ -91,6 +91,18 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>
     Ok(organization)
 }
 
+pub async fn find_by_external_id(
+    db: &DatabaseConnection,
+    external_id: &ExternalId,
+) -> Result<Option<Model>, Error> {
+    let organization = Entity::find()
+        .filter(Column::ExternalId.eq(Uuid::parse_str(external_id).unwrap()))
```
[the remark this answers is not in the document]
jhodapp: Thanks, this is perfect.

entity_api/src/organization.rs
jhodapp: That makes sense, thanks for explaining it...it wasn't super obvious from what it looks like at first but now it does.

jhodapp (approved):

PR #48, opened 2024-05-21
Move away from external

calebbourg:
## Description
This PR removes the concept of an External ID and uses a tables primary key as the primary method of querying from the database.

### Basic thought process:

Coaching Relationships have a `coach_id` and a `coachee_id`. Both of which are needed to dispatch a subsequent request to fetch the Coach's user data and the Coachee's user data (for displaying their names, etc.). The issue is that `coach_id` and `coachee_id` are both, by design, primary keys on the users table. This means that in order to serve the `external_id` of each user I will need to do something like this:
for each `coaching_relationship` associated with the `organization`, fetch the `user` with primary key `user_id` _and_ the `user `with primary key `coachee_id` and _then_ serialize each users' `external_id` as the respective `coach_id` and `coachee_id` .

The reason for using External ID in the first place was to prevent the ability for bad actors to infer parts of the system based on the exposure of sequential integer primary keys. This PR solves that problem by using `UUID`s as primary keys.

**Note:** Integer primary keys are likely more performant than `UUID` strings. The trade off here is in the complexity of each database query. 
I believe it's a reasonable trade off

### Changes
* Remove External ID concept
* Use UUID primary keys

### Testing Strategy
1. Run the usual `cargo test` at the source tree root.

### Concerns
None


jhodapp:
@calebbourg a couple of things before I get around to a full review. First, can you please generate a new PNG diagram that we have sitting under entity_api/? Second, it seems the tests aren't passing possibly because of needing to run `cargo fmt` or some other Clippy error (I didn't get check for sure).

Thanks for this!

jhodapp (approved):
Looks great, thanks Caleb!

I updated it with a couple of changes before approving, just FYI:
1. Use the type alias `Id` everywhere instead of `Uuid` directly
2. Fixed the broken test in router.rs `update_an_organization_specified_by_id`

PR #50, opened 2024-06-27
Add coaching sessions endpoint

calebbourg:
## Description
This PR adds the ability to fetch coaching sessions by:
- `coaching_relationship_id`
- `from_date`
- `to_date`

### Changes
* Adds `GET /coaching_sessions` endpoint
* Adds `coaching_session::find_by` with the ability to filter by the above query params
* Adds seeds for `coaching_sessions` records
* Adds rapidoc specification for `GET /coaching_sessions`

### Testing Strategy
Tested locally using rapidoc
Screen Shot 2024-07-02 at 8 58 50 AM

### Concerns
I may consider moving the `coaching_relationship_id` into an explicit path parameter (`GET /coaching_relationships/:coaching_relationship_id/coaching_sessions`) so that this endpoint will require scoping the list of `coaching_sessions` to a specific `coaching_relationship_id`.  I think we can brainstorm on this when we implement an authorizations strategy.

### DIFF
--- a/entity/src/coaching_sessions.rs
+++ b/entity/src/coaching_sessions.rs
@@ -3,8 +3,9 @@
 use crate::Id;
 use sea_orm::entity::prelude::*;
 use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
 
-#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize, ToSchema)]
 #[sea_orm(schema_name = "refactor_platform", table_name = "coaching_sessions")]
 pub struct Model {
     #[sea_orm(primary_key)]
--- a/entity/src/coaching_sessions.rs
+++ b/entity/src/coaching_sessions.rs
@@ -11,8 +11,8 @@ pub struct Model {
     #[sea_orm(primary_key)]
     pub id: Id,
     pub coaching_relationship_id: Id,
-    pub date: Option<DateTime>,
-    pub timezone: Option<String>,
+    pub date: DateTime,
+    pub timezone: String,
     pub created_at: DateTimeWithTimeZone,
     pub updated_at: DateTimeWithTimeZone,
 }
--- a/entity_api/src/coaching_session.rs
+++ b/entity_api/src/coaching_session.rs
@@ -1,0 +1,78 @@
+use super::error::{EntityApiErrorCode, Error};
+use crate::uuid_parse_str;
+use entity::{
+    coaching_sessions::{self, Entity, Model},
+    Id,
+};
+use sea_orm::{entity::prelude::*, DatabaseConnection};
+use std::collections::HashMap;
+
+pub async fn find_by_coaching_relationship(
+    db: &DatabaseConnection,
+    coaching_relationship_id: Id,
+) -> Result<Vec<Model>, Error> {
+    let query = by_coaching_relationship(Entity::find(), coaching_relationship_id).await;
+
+    Ok(query.all(db).await?)
+}
+
+pub async fn find_by(
+    db: &DatabaseConnection,
+    params: HashMap<String, String>,
+) -> Result<Vec<Model>, Error> {
+    let mut query = Entity::find();
+
+    for (key, value) in params {
+        match key.as_str() {
+            "coaching_relationship_id" => {
+                let coaching_relationship_id = uuid_parse_str(&value)?;
+                query = by_coaching_relationship(query, coaching_relationship_id).await;
+            }
+            _ => {
+                return Err(Error {
+                    inner: None,
+                    error_code: EntityApiErrorCode::InvalidQueryTerm,
+                });
+            }
+        }
+    }
+
+    Ok(query.all(db).await?)
+}
+
+async fn by_coaching_relationship(
+    query: Select<Entity>,
+    coaching_relationship_id: Id,
+) -> Select<Entity> {
+    query.filter(coaching_sessions::Column::CoachingRelationshipId.eq(coaching_relationship_id))
+}
+
+#[cfg(test)]
+// We need to gate seaORM's mock feature behind conditional compilation because
+// the feature removes the Clone trait implementation from seaORM's DatabaseConnection.
+// see https://github.com/SeaQL/sea-orm/issues/830
+#[cfg(feature = "mock")]
+mod tests {
+    use super::*;
+    use sea_orm::{DatabaseBackend, MockDatabase, Transaction};
+
+    #[tokio::test]
+    async fn find_by_coaching_relationships_returns_all_records_associated_with_coaching_relationship(
+    ) -> Result<(), Error> {
+        let db = MockDatabase::new(DatabaseBackend::Postgres).into_connection();
+
+        let coaching_relationship_id = Id::new_v4();
+        let _ = find_by_coaching_relationship(&db, coaching_relationship_id).await;
+
+        assert_eq!(
+            db.into_transaction_log(),
+            [Transaction::from_sql_and_values(
+                DatabaseBackend::Postgres,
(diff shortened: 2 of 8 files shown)


entity/src/coaching_sessions.rs
```diff
@@ -3,15 +3,16 @@
...
 
-#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize, ToSchema)]
 #[sea_orm(schema_name = "refactor_platform", table_name = "coaching_sessions")]
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: Id,
     pub coaching_relationship_id: Id,
-    pub date: Option<DateTime>,
-    pub timezone: Option<String>,
```
jhodapp, line 14: Great call, these shouldn't be optional.

entity_api/src/coaching_session.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub async fn find_by(
+    db: &DatabaseConnection,
+    params: HashMap<String, String>,
+) -> Result<Vec<Model>, Error> {
+    let mut query = Entity::find();
+
+    for (key, value) in params {
+        match key.as_str() {
+            "coaching_relationship_id" => {
```
jhodapp, line 15: @calebbourg I'm curious if there's a way of getting the key name as a string to compare to here, i.e. "coaching_relationship_id", "from_date", etc by adding these as constant string literals in entity/src/coaching_sessions.rs and referring to them here by constant name. If this is tricky and thus takes a lot of time, don't worry about it then. I was just thinking if it's straightforward we might as well make this change.

entity_api/src/coaching_session.rs
```diff
@@ -0,0 +1,121 @@
...
+            [Transaction::from_sql_and_values(
+                DatabaseBackend::Postgres,
+                r#"SELECT "coaching_sessions"."id", "coaching_sessions"."coaching_relationship_id", "coaching_sessions"."date", "coaching_sessions"."timezone", "coaching_sessions"."created_at", "coaching_sessions"."updated_at" FROM "refactor_platform"."coaching_sessions" WHERE "coaching_sessions"."date" < $1"#,
+                [to_date.into()]
+            )]
+        );
+
+        Ok(())
+    }
+}
```
jhodapp, line 121: One suggestion for another test is to test for a date range that doesn't match any coaching sessions and should return and empty `vec<Model>`. This is something that a user can run into on the frontend, so it'd be good to test for it returning cleanly.

entity_api/src/lib.rs
```diff
@@ -16,6 +17,13 @@ pub(crate) fn uuid_parse_str(uuid_str: &str) -> Result<Id, error::Error> {
     })
 }
 
+pub(crate) fn naive_date_parse_str(date_str: &str) -> Result<chrono::NaiveDate, error::Error> {
+    chrono::NaiveDate::parse_from_str(date_str, "%Y-%m-%d").map_err(|_| error::Error {
```
jhodapp, line 21: I'm curious if there's a specific reason you chose `NaiveDate` over `DateTime`? It seems that if you were to use `DateTime` instead, you would have a [more flexible string parser](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.parse_from_rfc3339) to call, and YYY-MM-DD is a valid ISO8601 date. The reason I noticed this is because I was working with the same date format on the frontend and was able to use the `ts-luxmon` library's `DateTime` type to parse ISO8601 dates.

jhodapp: Although, I did just notice you're only using this locally for tests, so perhaps this doesn't matter. Your call on this.

web/src/controller/coaching_session_controller.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use log::*;
+
+#[utoipa::path(
+    get,
+    path = "/coaching_sessions",
+    params(
+        ApiVersion,
+        ("coaching_relationship_id" = Option<Id>, Query, description = "Filter by coaching_relationship_id"),
+        ("from_date" = Option<NaiveDate>, Query, description = "Filter by from_date"),
```
jhodapp, line 22: Can we use `DateTime` instead of `NaiveDate`?

web/src/controller/coaching_session_controller.rs
```diff
@@ -0,0 +1,53 @@
...
+        (status = 405, description = "Method not allowed")
+    ),
+    security(
+        ("cookie_auth" = [])
+    )
+)]
+pub async fn index(
+    CompareApiVersion(_v): CompareApiVersion,
+    AuthenticatedUser(_user): AuthenticatedUser,
+    // TODO: create a new Extractor to authorize the user to access
```
jhodapp, line 37: Is this comment still relevant?

jhodapp (changes requested):
This looks so good Caleb! I left some questions and suggestions inline, take or leave any of them as you see fit.

entity_api/src/coaching_session.rs
calebbourg: That's a good idea. I'll give it a try

calebbourg: So for each `entity` the `DeriveEntityModel` adds a `Column` enum that has a variant for each of the table columns.  An example would be `coaching_relationship::Column::CoachingRelationshipId`. This, in my mind would be the best candidate to convert into strings or `&str` in order to compare in these `entity_api` functions. Unfortunately, there isn't a good built in trait for converting enums to `&str` or `&'static str`. There are a few crates that provide implementations like [strum](https://github.com/Peternator7/strum) though I think for now I'd like to hold off on adding the dependency and see if this pattern that we're using here stabilizes in our project.

entity_api/src/coaching_session.rs
calebbourg: We can't really test that here using this methodology as we're not actually testing what the query returns in terms of records. We're testing the SQL generated by the query.

entity_api/src/lib.rs
calebbourg: I chose `NaiveDate` because `DateTime` is "timezone aware" as stated in their [docs](https://docs.rs/chrono/latest/chrono/). In our case we chose to set the timezone as a separate column on the table as opposed to encoded in the actual timestamps. We should definitely come back and revisit how this is set up and make sure it's what we're looking for. I'd propose considering that worthy of a discussion and a separate PR if that works for you.

web/src/controller/coaching_session_controller.rs
calebbourg: I have been copying these over from the other controllers as I believe it is relevant since we currently have no concept of authorization. Made sense to me to propagate the comment just so we would have a sense of where we might want authorization. Open to your thoughts there.

entity_api/src/coaching_session.rs
jhodapp: Sounds good Caleb, thanks for looking into it it a little bit.

entity_api/src/coaching_session.rs
jhodapp: More a test for the controller level then? I don't want to get hung up on tests but I did want to ask about anything that came up as a question in my mind while reviewing.

entity_api/src/lib.rs
jhodapp: Ok sounds good. I created a [discussion to bookmark this conversation](https://github.com/Jim-Hodapp-Coaching/refactor-platform-rs/discussions/51#discussion-6901596) and discuss at the right time.

web/src/controller/coaching_session_controller.rs
jhodapp: Bookmarked as a [future discussion](https://github.com/Jim-Hodapp-Coaching/refactor-platform-rs/discussions/51#discussion-6901596).

web/src/controller/coaching_session_controller.rs
jhodapp: Ok, so it does still apply. Very good.

jhodapp (approved):
Looks great Caleb, very nice work as always!

PR #57, opened 2024-07-27
Retrieve a Note by its id

jhodapp:
## Description
This PR adds the ability to retrieve a Note by its id.

### Changes
* Retrieve a Note by its id
* Make sure to return an error when an invalid id is specified

### Testing Strategy
1. Using [RAPIdoc](http://localhost:4000/rapidoc#get-/notes/-id-), try and retrieve an existing Note by its Id (look in the DB for an existing Note's id to test)
2. Observe that a 200 is returned with a Note JSON object 
3. Try entering a valid Id but of a Note that doesn't yet exist
4. Observe that a 204 is returned and No Content

### Concerns
None

### DIFF
--- a/entity_api/src/note.rs
+++ b/entity_api/src/note.rs
@@ -57,6 +57,32 @@ pub async fn update(db: &DatabaseConnection, id: Id, model: Model) -> Result<Mod
     }
 }
 
+pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>, Error> {
+    match Entity::find_by_id(id).one(db).await {
+        Ok(Some(note)) => {
+            debug!("Note found: {:?}", note);
+
+            Ok(Some(note))
+        }
+        Ok(None) => {
+            debug!("Note with id {} not found", id);
+
+            Err(Error {
+                inner: None,
+                error_code: EntityApiErrorCode::RecordNotFound,
+            })
+        }
+        Err(err) => {
+            debug!("Note with id {} not found and returned error {}", id, err);
+            Err(Error {
+                inner: None,
+                error_code: EntityApiErrorCode::RecordNotFound,
+            })
+            
+        }
+    }
+}
+
 pub async fn find_by(
     db: &DatabaseConnection,
     query_params: HashMap<String, String>,
--- a/entity_api/src/note.rs
+++ b/entity_api/src/note.rs
@@ -47,7 +47,7 @@ pub async fn update(db: &DatabaseConnection, id: Id, model: Model) -> Result<Mod
             Ok(active_model.update(db).await?.try_into_model()?)
         }
         None => {
-            debug!("Note with id {} not found", id);
+            error!("Note with id {} not found", id);
 
             Err(Error {
                 inner: None,
@@ -65,7 +65,7 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>
             Ok(Some(note))
         }
         Ok(None) => {
-            debug!("Note with id {} not found", id);
+            error!("Note with id {} not found", id);
 
             Err(Error {
                 inner: None,
@@ -73,7 +73,7 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>
             })
         }
         Err(err) => {
-            debug!("Note with id {} not found and returned error {}", id, err);
+            error!("Note with id {} not found and returned error {}", id, err);
             Err(Error {
                 inner: None,
                 error_code: EntityApiErrorCode::RecordNotFound,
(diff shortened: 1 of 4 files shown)


qafui (approved):
Nice work!

entity_api/src/note.rs
```diff
@@ -79,7 +104,27 @@ pub async fn find_by(
         }
     }
 
-    Ok(query.all(db).await?)
+    match query.all(db).await {
```
calebbourg, line 107: Disclaimer:  This is from own standards in my brain which I never put on paper so take it with a grain of salt.

It was an intentional decision to return the empty collection when a query that results in a `WHERE`  SQL clause returns 0 records. This is conventional in other ORMs that I've used. 

`RecordNotFound` should only be returned for queries that result in a `SELECT` SQL statement.

The reason being is exactly so that we didn't have to do the extra check like `!notes.is_empty()`

I'm curious what the reasoning here is and if we could provide the same benefits without the extra nesting and conditional logic.

entity_api/src/note.rs
```diff
@@ -79,7 +104,27 @@ pub async fn find_by(
...
+            if !notes.is_empty() {
+                Ok(notes)
+            } else {
+                Err(Error {
+                    inner: None,
+                    error_code: EntityApiErrorCode::RecordNotFound,
+                })
+            }
+        }
+        Err(err) => {
```
calebbourg, line 120: Similarly to above. when an `Err()` occurs from `query.all(db).await`, it should be propagated directly to the caller to be converted to a top level error.

entity_api/src/note.rs
```diff
@@ -57,6 +57,31 @@ pub async fn update(db: &DatabaseConnection, id: Id, model: Model) -> Result<Mod
     }
 }
 
+pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Option<Model>, Error> {
+    match Entity::find_by_id(id).one(db).await {
```
calebbourg, line 61: This makes sense here I think

calebbourg (changes requested):
This looks great! Some of the changes bring up some great points about our conventions that might be good to discuss at this point.

I took a look at the [JSON API specification for fetching resources](https://jsonapi.org/format/#fetching-resources-responses-200) and it outlines as convention:

- Always returning  an HTTP `200` and an empty list from endpoints that return collections when the collection is empty
- Always returning an HTTP `200` and  `NULL` from endpoints that return a single record when the record doesn't exist.

I think that adopting these standards makes sense considering their source as well as allowing us to simplify the BE code here and remove some of the nesting of `match` and other conditionals.

This makes me think we might want to update the code in our `web` crate to convert the `RecordNotFound` error to a `200`  `{"data": NULL}`

entity_api/src/note.rs
jhodapp: @calebbourg You wouldn't ever get an error for the empty case, so you'd always get a 200 OK which is not correct, the record isn't found. So that second match branch, i.e. `Err(err)` does not occur when a `Note` isn't found, for example, if you provide an id of a Note that doesn't exist.

calebbourg: I'm proposing that we do actually want to return a `200` when no records exist for index fetches to align with the [JSON API convention](https://jsonapi.org/format/#fetching-resources-responses-200).

> server MUST respond to a successful request to fetch a resource collection with an array of [resource objects](https://jsonapi.org/format/#document-resource-objects) or an empty array ([]) as the response document’s primary data.

I think the philosophy is that the request itself is still "successful" in that there were no exceptions or auth issues. We just happened to return 0 records.

Reading more into the spec I think that the [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) we're mapping to `RecordNotFound` is not semantically inline with the HTTP standard as it states.

> This might be used, for example, when implementing "save and continue editing" functionality for a wiki site. In this case a [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) request would be used to save the page, and the 204 No Content response would be sent to indicate that the editor should not be replaced by some other page.


Which I read as "something was successful here but this endpoint itself doesn't return anything meaningful". The fact that adhering to the standard yields less matching and conditional logic in the code leads me to believe that seaORM may have designed their API to support these patterns. 

This is all interpretation, of course, and I'm learning these standards as we go.

jhodapp: Now having read the JSON API spec in this area, here's what I see it saying:



https://jsonapi.org/format/#fetching-resources-responses-404

There are 3 cases:
1) Return `[]` for a resource that does exist, e.g. http://localhost/articles/ but there are no articles yet but also the URI exists.
2) Return a `null` for an nested entity that doesn’t exist (e.g. an article that has an author that hasn’t been defined yet) under the requested resource
3) A request for an instance of an entity at a specific URI that doesn’t exist should return a 404

calebbourg (approved):
Looks great!!

PR #66, opened 2024-09-25
Update statuses

calebbourg:
## Description
This PR adds two new endpoings:
- `PUT actions/:id/status`
- `PUT overarching_goals/:id/status`

As well as the API plumbing to utilize the endpoints for updating `status` columns

### Changes
* Adds `PUT actions/:id/status`
* Adds `PUT overarching_goals/:id/status`


entity/src/status.rs
```diff
@@ -19,3 +19,15 @@ impl std::default::Default for Status {
         Self::InProgress
     }
 }
+
+impl From<&str> for Status {
```
jhodapp, line 23: Nice addition!

jhodapp (changes requested):
Looking great, thanks for this! One change needed is to add the new controller methods to the openapi path macro in router.rs.

jhodapp (approved):
Looks fantastic, thank you!

PR #67, opened 2024-09-26
Improve Action update behavior, and add the ability to delete an action

jhodapp:
## Description
This PR improves Actions update behavior, and add the ability to delete an action.

**NOTE** Make sure to rerun the scripts/rebuild_db.sh script and re-seed the DB with cargo run --bin seed_db

### Changes
* Improve Action UPDATE behavior
* Add ability to DELETE an Action by its Id

### Testing Strategy
1. Make sure to rerun the `scripts/rebuild_db.sh` script and re-seed the DB with `cargo run --bin seed_db`
2. Using RAPidoc, add an Action and then try and delete it

### Concerns
None


docs/db/refactor_platform_rs.dbml
```diff
@@ -81,9 +81,9 @@ Table refactor_platform.actions {
   user_id uuid [not null, note: 'User that created (owns) the action']
   due_by timestamptz
   status status [not null]
-  status_changed_at timestamptz
-  created_at timestamp [not null, default: `now()`]
-  updated_at timestamp [not null, default: `now()`]
+  status_changed_at timestamptz [not null, default: `now()`]
```
calebbourg, line 84: Just to be clear, I think this will set `status_changed_at` to "now" when we first create an action. Which might be somewhat misleading for newly created actions. TBH that seems reasonable to me but want to make sure that was your intention.

web/src/controller/action_controller.rs
```diff
@@ -146,3 +144,32 @@ pub async fn index(
...
+/// DELETE an Action specified by its primary key.
+#[utoipa::path(
+    delete,
+    path = "/actions/{id}",
+    params(
+        ApiVersion,
+        ("id" = i32, Path, description = "Action id to delete")
+    ),
+    responses(
+        (status = 200, description = "Successfully deleted a certain Action by its id", body = [i32]),
```
calebbourg, line 157: I read the function as returning
```
{
  "data": {
    "id": {id}
  }
}
```
If that is the case maybe we could either:
- Define a data structure that reflects that and use it as a schema for `body = ` here or
- Just return the full deleted object

I want to say I've seen the latter but I don't have strong feelings about it.

docs/db/refactor_platform_rs.dbml
jhodapp: Great question to double check. Yes, this is intentional. It's because we don't have a non-state option for `Status`. If we did, I'd leave this as something we could leave unset. The way I tweaked things in this PR and the frontend one for actions is the "not_started" enum is chosen by default, so it is being set.

Thoughts?

web/src/controller/action_controller.rs
jhodapp: We can update this later if you're cool with that, but this is how all of our deletes work today, so I'm just copying them. I'd much rather return the entire thing which is what I do on the frontend anyway.

calebbourg (approved):

PR #79, opened 2024-12-02
Deserialize just the right fields when creating new entities

jhodapp:
## Description
This PR fixes each entity's SeaORM model such that we can call the create endpoints with just the right fields for creating a new instance of each one.

### Changes
* Ensure we deserialize just the right fields when creating new entities
* Make sure we can set a password when creating a new `User` but still not able to serialize it for returning an existing `User`'s hashed password
* Use `#[schema(value_type = String, format = DateTime)]` consistently on `DateTimeWithTimeStamp` entity fields

### Testing Strategy
1. From the RAPIdoc UI (http://localhost:4000/rapidoc)
2. Sign in with a test user
3. Try creating a new instance of each entity. Check over the entity's schema
4. Ensure that the schema only includes fields that you expect (i.e. they don't include things like `id` or `created_at`
5. Each entity should create successfully

### Concerns
None

### DIFF
--- a/entity/src/actions.rs
+++ b/entity/src/actions.rs
@@ -20,8 +20,10 @@ pub struct Model {
     #[serde(skip_deserializing)]
     pub status_changed_at: DateTimeWithTimeZone,
     #[serde(skip_deserializing)]
+    #[schema(value_type = String, format = DateTime)] // Applies to OpenAPI schema
     pub created_at: DateTimeWithTimeZone,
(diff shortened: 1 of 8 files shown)


jhodapp:
I totally agree with you that we'll want to take a broader look at user creation. For now, I just decided to fix what exists, and am letting the client application I'm writing to create production entities for my clients handle pulling everything together via our basic API.

PR #85, opened 2024-12-11
select distinct organizations

calebbourg:
## Description
Fixes an issue where the BE was serving 2 copies of the same `organizations` records

### Changes
* add `DISTINCT` to the SQL `SELECT` statements

#### Tested Locally
Screen Shot 2024-12-11 at 2 15 34 PM

### Concerns


jhodapp (approved):
LGTM

PR #89, opened 2024-12-16
Fix overarching goals and actions not saving

jhodapp:
## Description
This PR fixes overarching goals and actions not saving by put custom enum type `status` is in the `refactor_platform` DB schema instead of the `public` default one.

**Note** PR #88 must land before this one does.

### Changes
* Fix overarching goals and actions not saving

### Testing Strategy
1. Log in and join a session
2. Ensure that both an OverarchingGoal and at least 1 Action can be created successfully

### Concerns
None


calebbourg (approved):
Very interesting. Nice catch.

PR #98, opened 2025-01-31
Standardize coaching session date

calebbourg:
## Description
This PR updates `coaching_sessions` creation to truncate seconds on the `date` column. The reasoning behind this is that Coaching Sessions will be scheduled on a minute bases (Ex. Feb 1st at 10:30 **not** Feb 1st at 10:30:25). Eventually we will validate that a Coaching Session isn't scheduled twice at the same time for a given coaching relationship. This will be easier to enforce without the addition of seconds and subseconds.

This PR also adds `coaching_sessions.collab_document_name` which stores our Tiptap document names. I intentionally chose to not include Tiptap in the naming to future proof against vendor lock should the need ever arise to move away from Tiptap.

### Changes
* Truncates `date` seconds
* Adds `coaching_sessions.collab_document_name`

#### Tested using Utoipa
Screenshot 2025-01-31 at 10 32 05 AM


service/src/config.rs
```diff
@@ -75,7 +75,7 @@ pub struct Config {
         short,
         long,
         env,
-        default_value_t = LevelFilter::Warn,
```
calebbourg, line 78: Changing this as I always wind up changing it when developing locally. We want to be able to see `Info` level logs as a default.

jhodapp: This is intentionally set to Warn as the default so we don't unnecessarily expose details for a production run, and then I always run the backend with the following command line: `cargo run  -- -l ERROR -d postgres://refactor:password@localhost:5432/refactor_platform --allowed-origins="http://localhost:3000"`

jhodapp (changes requested):
One minor change to consider.

service/src/config.rs
jhodapp: Discussed and agreed that Warn makes sense even for production runtime.

jhodapp (approved):
LGTM

PR #102, opened 2025-02-24
Refactor parameter typing

calebbourg:
## Description
This PR refactors how we type incoming parameters for certain types of requests.

It adds a `QueryFilterMap` for type safety around parameters used to filter database queries.
It adds a related `IntoQueryFilterMap` trait which provides the mechanism for translating types between application layers.

This way we can provide types for endpoint parameters in the `web` layer and then translate that data into types understood and used by `domain` and `entity_api` without these lower layers needing to depend on `web` (avoiding circular dependency)

This a step toward my current goal of removing `web` dependency on anything other than `domain`

### Changes
* Adds `QueryFilterMap`
* Adds `IntoQueryFilterMap`
* Updates `web::error::Error`

### Testing Strategy
I tested this via Rapidoc and against the FE

### Next
- Apply this pattern to other endpoints
- Re-Export any necessary `entity_api` or `entity` types through `domain`
- Remove `web` dependencies on `entity_api` and `entity`
- Add data validation mechanism to `domain`
- Allow for partial updates to data (for user, coaching session, etc. updates)

### DIFF
--- a/web/src/controller/agreement_controller.rs
+++ b/web/src/controller/agreement_controller.rs
@@ -2,16 +2,16 @@ use crate::controller::ApiResponse;
 use crate::extractors::{
     authenticated_user::AuthenticatedUser, compare_api_version::CompareApiVersion,
 };
+use crate::params::agreement::IndexParams;
 use crate::{AppState, Error};
 use axum::extract::{Path, Query, State};
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 use axum::Json;
+use domain::agreement as AgreementApi;
 use entity::{agreements::Model, Id};
-use entity_api::agreement as AgreementApi;
 use serde_json::json;
 use service::config::ApiVersion;
-use std::collections::HashMap;
 
 use log::*;
 
@@ -122,7 +122,7 @@ pub async fn update(
     path = "/agreements",
     params(
         ApiVersion,
-        ("coaching_session_id" = Option<Id>, Query, description = "Filter by coaching_session_id")
+        ("coaching_session_id" = Id, Query, description = "Filter by coaching_session_id")
     ),
     responses(
         (status = 200, description = "Successfully retrieved all Agreements", body = [entity::agreements::Model]),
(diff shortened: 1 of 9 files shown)


web/src/controller/agreement_controller.rs
```diff
@@ -139,15 +139,12 @@ pub async fn index(
     // TODO: create a new Extractor to authorize the user to access
     // the data requested
     State(app_state): State<AppState>,
-    Query(params): Query<HashMap<String, String>>,
+    // params: Result<Query<IndexParams>, QueryRejection>,
```
jhodapp, line 142: @calebbourg Is this needed as documentation?

web/src/error.rs
```diff
@@ -28,41 +37,100 @@ impl std::fmt::Display for Error {
...
-                        warn!("ExternalErrorKind::Network: Responding with 502 Bad Gateway. Error: {:?}", self);
-                        (StatusCode::BAD_GATEWAY, "BAD GATEWAY").into_response()
-                    }
-                    ExternalErrorKind::Other => {
-                        warn!("ExternalErrorKind::Other: Responding with 500 Internal Server Error. Error: {:?}", self);
-                        (StatusCode::INTERNAL_SERVER_ERROR, "INTERNAL SERVER ERROR").into_response()
-                    }
-                }
+        match self {
+            Error::Domain(ref domain_error) => self.handle_domain_error(domain_error),
```
jhodapp, line 41: This is nice, I didn't know how to use `ref` like this in such a way as to examine it for pattern matching without consuming/moving it.

jhodapp (changes requested):
Just one minor thing to adjust, this looks fantastic. Nice evolution of our new `domain` layer on top of `entity_api` and encapsulating types that really don't need to be exposed.

web/src/error.rs
calebbourg: I didn't know about it either!

web/src/controller/agreement_controller.rs
calebbourg: Nope :) updated

jhodapp (approved):

PR #103, opened 2025-02-25
More refactoring

calebbourg:
## Description

This PR accomplishes the following:
- Removes `web` crate _direct_ dependency on `entity_api`
- Removes `web` crate _direct_ dependency on `entity`
- Removes `domain` crate _direct_ dependency on `entity`

### The current dependency graph for these crates is as follows:
```mermaid
graph TD;
    web-->domain;
    web-->service;
    web-->entity_api
    web-->entity
    domain-->entity_api;
    domain-->entity
    entity_api-->entity;
    entity_api-->service;
```

### The new dependency graph for these crates is as follows:
```mermaid
graph TD;
    web-->domain;
    web-->service;
    domain-->entity_api;
    entity_api-->entity;
    entity_api-->service;
```

I use _direct_ here because we are using re-exports to accomplish this. For example `domain` has access to `entity` types via re-export through `entity_api`. This means there is still some amount of coupling.

My hypothesis for this being beneficial regardless of the indirect coupling is that we'll now have to explicitly re-add crates to the various `Cargo.toml` files in order to re-introduce these dependencies. AKA we will not be able to do it casually. My hope is that this will force us to make any decisions like that with an intentional mindset and we will be able to flag things like that in PRs.

This PR also extracts `find_by` filtering into a generic function and updates the various entity related systems to use the new function. This is a continuation of the new pattern introduced in #102 

**Note:** There are a lot of changed files here, but it's the same changes over and over. I split things up logically into commits so reviewing commit by commit might be helpful.

### Changes
* Remove direct crate dependencies
* Define and use new `find_by` functino

### Testing Strategy
I tested our current functionality locally against the FE

### DIFF
--- a/docs/architecture/crate_dependency_graph.md
+++ b/docs/architecture/crate_dependency_graph.md
@@ -1,0 +1,11 @@
+
+This diagram represents the dependency structure of the crates in this project. Each arrow indicates a dependency relationship between the crates. For example, the `web` crate depends on both the `domain` and `service` crates, while the `entity_api` crate depends on the `entity` and `service` crates.
+
+```mermaid
+graph TD;
+    web-->domain;
+    web-->service;
+    domain-->entity_api;
+    entity_api-->entity;
+    entity_api-->service;
+```
--- a/domain/src/jwt/mod.rs
+++ b/domain/src/jwt/mod.rs
@@ -13,7 +13,7 @@
 //! use domain::jwt::generate_collab_token;
 //! use sea_orm::DatabaseConnection;
 //! use service::config::Config;
-//! use entity::Id;
+//! use crate::Id;
 //!
 //! async fn example(db: &DatabaseConnection, config: &Config, coaching_session_id: Id) {
 //!     match generate_collab_token(db, config, coaching_session_id).await {
@@ -23,17 +23,13 @@
 //! }
 //! ```
 
-use crate::coaching_session;
 use crate::error::{DomainErrorKind, Error, InternalErrorKind};
+use crate::{coaching_session, jwts::Jwts, Id};
 use claims::TiptapCollabClaims;
-use entity::Id;
 use jsonwebtoken::{encode, EncodingKey, Header};
 use log::*;
 use sea_orm::DatabaseConnection;
 use service::config::Config;
-
-// re-export the Jwt struct from the entity module
-pub use entity::jwt::Jwt;
 
 pub(crate) mod claims;
 
@@ -47,7 +43,7 @@ pub async fn generate_collab_token(
     db: &DatabaseConnection,
     config: &Config,
     coaching_session_id: Id,
-) -> Result<Jwt, Error> {
+) -> Result<Jwts, Error> {
     let coaching_session = coaching_session::find_by_id(db, coaching_session_id).await?;
 
     let collab_document_name = coaching_session.collab_document_name.ok_or_else(|| {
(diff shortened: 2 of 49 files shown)


entity/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub mod coachees;
 pub mod coaches;
 pub mod coaching_relationships;
 pub mod coaching_sessions;
-pub mod jwt;
+pub mod jwts;
```
calebbourg, line 11: For consistency with other naming conventions

jhodapp (changes requested):
I want to digest this some more, but I have one inline question and I'm also wondering if you could take your diagram in this PR and add that under a new directory/file pair /docs/systems/design.md ?

domain/src/jwt/mod.rs
```diff
@@ -47,7 +43,7 @@ pub async fn generate_collab_token(
     db: &DatabaseConnection,
     config: &Config,
     coaching_session_id: Id,
-) -> Result<Jwt, Error> {
+) -> Result<Jwts, Error> {
```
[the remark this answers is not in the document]
calebbourg: It felt a bit strange for me as well, but this is the convention for all modules in the `entity` crate and a convention we inherited from the seaORM examples. I think keeping with the convention is probably best. It makes more sense when you consider that database table names are conventionally plural and most entity data structures here map directly to a table. This struct represents a JWT which is one of potentially multiple types of JWTs.

calebbourg:
@jhodapp diagram added. Great suggestion. Github renders the markdown as well!

Screenshot 2025-02-25 at 9 07 38 PM

domain/src/jwt/mod.rs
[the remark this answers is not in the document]
calebbourg: Just thinking about this a bit more.. I changed it so that the struct itself is `Jwt` (singular) and kept the enclosing module as `jwts` (plural) to match the convention. Similar to how `actions` is plural but `actions::Entity` is singular.
Let me know what you think.
2d0b5871ae7c8336eeaeffcc689e6c71a3782c13

docs/architecture/crate_dependency_graph.md
```diff
@@ -0,0 +1,11 @@
+
```
jhodapp, line 1: Beautiful, love it!

entity_api/src/query.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{error::Error, QueryFilterMap};
+use sea_orm::strum::IntoEnumIterator;
+use sea_orm::{ColumnTrait, DatabaseConnection, EntityTrait, QueryFilter};
+
+/// Find all records of an entity by the given query filter map.
+pub async fn find_by<E, C>(
```
jhodapp, line 6: This is some really clean, readable code, and it's in the style of how SeaORM approaches working with entities as well.

entity_api/src/agreement.rs
```diff
@@ -63,62 +62,15 @@ pub async fn update(db: &DatabaseConnection, id: Id, model: Model) -> Result<Mod
...
-    for column in agreements::Column::iter() {
-        if let Some(value) = query_filter_map.get(&column.to_string()) {
-            query = query.filter(column.eq(value));
-        }
-    }
-
-    Ok(query.all(db).await?)
+pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Model, Error> {
+    Entity::find_by_id(id).one(db).await?.ok_or_else(|| Error {
+        source: None,
```
jhodapp, line 71: I love seeing this simplification of the tests that `Entity::find_by_id()` provides. This should make our tests even more reliable in the long run. Nicely done.

jhodapp (approved):
Looks fantastic.

PR #105, opened 2025-03-03
Make sure that creating a new Organization accepts the "slug"  field

jhodapp:
## Description
This PR ensures that when creating a new Organization instance, it accepts the `slug` field and this value is unique across all Organizations.

### Changes
* Accept the `slug` field as non-optional
* Ensure `slug`'s value is unique across all Organizations

### Testing Strategy
1. Create a new organization with RAPIdoc

### Concerns
None

### DIFF
--- a/entity/src/organizations.rs
+++ b/entity/src/organizations.rs
@@ -15,7 +15,7 @@ pub struct Model {
     #[sea_orm(unique)]
     pub name: String,
     pub logo: Option<String>,
-    #[serde(skip_deserializing)]
+    #[sea_orm(unique)]
     pub slug: String,
     #[serde(skip_deserializing)]
     #[schema(value_type = String, format = DateTime)] // Applies to OpenAPI schema
--- a/entity_api/src/organization.rs
+++ b/entity_api/src/organization.rs
@@ -19,8 +19,9 @@ pub async fn create(db: &DatabaseConnection, organization_model: Model) -> Resul
     let now = Utc::now();
 
     let organization_active_model: ActiveModel = ActiveModel {
-        logo: Set(organization_model.logo),
         name: Set(organization_model.name),
+        logo: Set(organization_model.logo),
+        slug: Set(organization_model.slug),
         created_at: Set(now.into()),
         updated_at: Set(now.into()),
         ..Default::default()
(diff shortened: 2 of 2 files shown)


entity/src/organizations.rs
```diff
@@ -15,7 +15,7 @@ pub struct Model {
     #[sea_orm(unique)]
     pub name: String,
     pub logo: Option<String>,
-    #[serde(skip_deserializing)]
+    #[sea_orm(unique)]
```
calebbourg, line 18: First of all, awesome find on the `unique` macro! Without looking at the docs, I wonder if we can express something like `unique in the scope of {other associated record}`. We'll need something like that for coaching relationships. 

As far as generating the slugs on the FE, I have a fairly strong opinion that this should be a BE operation. The way I had envisioned it would be that a mechanism would exist to derive a `slug` by translating the `name` field automatically at the time of creation. The reasons for having it done of the BE:
- The `slug` is designed for programmability due to its strict standardized requirements, and its immutability. I'd prefer to use BE technology to handle the rules involved and the string manipulation.
-The generation of a `slug`  cannot be susceptible to savy user's modifying client code as we will be depending on characteristics of the `slug` (Ex. hyphenated, lowercase, etc.) in our downstream code.

We, of course, will add validations on the format of a `slug` on the BE but I'd like to keep this logic and responsibility out of the FE if possible to keep things simple and thin.

Let me know what you think about that.

calebbourg (changes requested):
I'd like us to generate `slug` values in the BE

jhodapp:
Oh I see. Something still needs to change then because creating a new Organization without providing a value for 'slug' was failing to insert claiming I still needed to set a value.

calebbourg (approved):

jhodapp:
> @jhodapp yeah that's right. We need to add a step in the creation process where the input `name` is translated into a `slug` in some deterministic way.
> 
> If this is a blocker for you at the moment (assuming you need it to seed orgs with the tool you built)I'm happy to approve this PR now and come back and move the `slug` creation to the BE later.

Thanks for that flexibility Caleb, but I do think we should do it right, right now. Or at the very least make it so that the frontend can create a new Organization without providing a slug and still have it insert into the DB successfully.

I'm happy to take a crack at this so you can keep moving forward on your feature work.

jhodapp:
Superseded by #107

PR #108, opened 2025-03-04
Add a partial update User endpoint

calebbourg:
## Description

This PR introduces a pattern for providing partial updates to database tables via our API.
By _partial_ updates I mean that we are able to modify a subset of columns for any given table without having to provide values for the the other columns. The pattern is somewhat based on the pattern for filtering queries as introduced in #102 

**Note:** This new endpoint assumes that the user being updated will always be the current user (as noted inline in comments). At some point we will likely want a user to be able to make updates to another user. We can decide how the code diverges at that point when we know more about how we want that to work.

### Changes
* Adds generic `update` function
* Updates user updating to use new function
* Adds `PUT /users` update endpoint
* Fixes a small issue with the Rapidoc for login

#### I was able to test partially updating a user in Rapidoc
Screenshot 2025-03-04 at 1 18 55 PM

### Coming Up
- Adding New Users
- Updating Coaching Sessions
- Stricter validations

### DIFF
--- a/web/src/controller/user_controller.rs
+++ b/web/src/controller/user_controller.rs
@@ -54,7 +54,7 @@ pub async fn create(
     ),
     request_body = UpdateUserParams,
     responses(
-        (status = 200, description = "Successfully updated a User", body = [users::Model]),
+        (status = 204, description = "Successfully updated a User", body = ()),
         (status = 401, description = "Unauthorized"),
     ),
     security(
@@ -67,6 +67,6 @@ pub async fn update(
     State(app_state): State<AppState>,
     Json(params): Json<UpdateUserParams>,
 ) -> Result<impl IntoResponse, Error> {
-    let updated_user = UserApi::update(app_state.db_conn_ref(), user.id, params).await?;
-    Ok(Json(ApiResponse::new(StatusCode::OK.into(), updated_user)))
+    UserApi::update(app_state.db_conn_ref(), user.id, params).await?;
+    Ok(Json(ApiResponse::new(StatusCode::NO_CONTENT.into(), ())))
 }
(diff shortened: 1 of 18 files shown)


web/src/params/user.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Debug, Deserialize, IntoParams, ToSchema)]
+pub struct UpdateUserParams {
+    pub email: Option<String>,
+    pub first_name: Option<String>,
+    pub last_name: Option<String>,
+    pub display_name: Option<String>,
+    pub github_profile_url: Option<String>,
+}
+
+impl IntoUpdateMap for UpdateUserParams {
```
jhodapp, line 16: I love this pattern for updating just the right fields - it's extremely clean, easy to understand and effective.

web/src/controller/user_controller.rs
```diff
@@ -36,3 +39,34 @@ pub async fn create(
...
+#[utoipa::path(
+    put,
+    path = "/users",
+    params(
+        ApiVersion,
+        UpdateUserParams
+    ),
+    request_body = UpdateUserParams,
+    responses(
+        (status = 200, description = "Successfully updated a User", body = [users::Model]),
```
jhodapp, line 57: What are your thoughts on starting to follow the spec on a successful PUT operation for what HTTP response status code we respond with? It seems like we can respond with a 204, and there's no body returned in a successful update. There's also a 201 response as well for when a `User` could be created in the process of doing a PUT.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

jhodapp (commented):
@calebbourg Looks really great, I like the new UpdateMap pattern and IntoUpdateMap trait a lot.

One question I'm curious for your thoughts on inline.

web/src/controller/user_controller.rs
calebbourg: Good call out. I went back and forth on this but the fact that you're bringing it up, I think, tips the scales. Update to return 204 in 67be87dc2173cb9da0864924750deaeb8084bd18

jhodapp (approved):

PR #111, opened 2025-03-23
88 feature add member management page

calebbourg:
## Description

This PR builds on the existing Users API for use with our members page and user (member) CRUD operations.

The most notable concept here is scoping indexing of users under `/organizations/{organization_id}/users`. This is as opposed to what we have historically done which is consider something like `organization_id` as a filter parameter.  Ex. `/users?organization_id={organization_id}`

The philosophy here is to move toward a model where filter parameters, AKA using query params, should be reserved for **fields that exist directly on the table being queried**. 
So for the `users` example, filtering by something like `display_name` will be done using query params (`/users?display_name={display_name}`) whereas, such as in this case, fields outside of `users` will be used for _scope_. 

The intention behind this is to formalize the difference between filtering and scope and to provide a framework for making the decision of when to do each. I'm finding (at least at the moment) that this distinction produces cleaner, more flexible, code. 

**Note:** this does not prevent us from using `/users` endpoints in the future for situations where it makes sense. The `/users/{user_id}/update` endpoint may fall into this category

### Changes
* Adds User API operations
* Index users under the scope of organizations.
* Adds `organizations_users`  join table with unique compound index for `user_id` and `organization_id`

### Testing Strategy
I tested this locally against https://github.com/refactor-group/refactor-platform-fe/pull/89

Screenshot 2025-03-23 at 2 33 19 PM
Screenshot 2025-03-23 at 2 33 14 PM

### Concerns
At some point, if this pattern works in our favor, we should come back and convert other similar endpoints to this style. Ex. `/coaching_sessions/{coaching_session}/{resource}`


jhodapp (approved):
Looks great, thanks Caleb.

PR #113, opened 2025-03-27
88 feature add member management page create

calebbourg:
## Description

This PR allows for Coaches to create new Coachees through the API.

The order of operations are as follows:
- A new user is created
- The new user is associated with the current organization
- A new coaching relationship is created where the new user is the coachee and the current authenticated user is the coach

### Changes
* Adds `POST /organizations/{organization_id}/users` endpoint

### Testing Strategy
Tested locally against https://github.com/refactor-group/refactor-platform-fe/pull/92

### DIFF
--- a/domain/src/user.rs
+++ b/domain/src/user.rs
@@ -1,12 +1,16 @@
-use crate::{error::Error, users, Id};
-use entity_api::{mutate, query, query::IntoQueryFilterMap};
-use sea_orm::DatabaseConnection;
-use sea_orm::IntoActiveModel;
-
+use crate::{
+    error::Error,
+    error::{DomainErrorKind, EntityErrorKind, InternalErrorKind},
+    users, Id,
+};
+use chrono::Utc;
 pub use entity_api::user::{
     create, create_by_organization, find_by_email, find_by_id, find_by_organization, AuthSession,
     Backend, Credentials,
 };
+use entity_api::{mutate, query, query::IntoQueryFilterMap};
+use sea_orm::IntoActiveModel;
+use sea_orm::{DatabaseConnection, TransactionTrait};
 
 pub async fn find_by(
     db: &DatabaseConnection,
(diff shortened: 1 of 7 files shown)


domain/src/user.rs
```diff
@@ -1,11 +1,16 @@
...
-
+use crate::{
+    error::Error,
+    error::{DomainErrorKind, EntityErrorKind, InternalErrorKind},
+    users, Id,
+};
+use chrono::Utc;
 pub use entity_api::user::{
-    create, find_by_email, find_by_id, find_by_organization, AuthSession, Backend, Credentials,
+    create, create_by_organization, find_by_email, find_by_id, find_by_organization, AuthSession,
```
calebbourg, line 8: `create_by_organization` will be the function we use once we decouple the user creation from coaching relationship assignment

web/src/protect/organizations/users.rs
```diff
@@ -17,7 +18,45 @@ pub(crate) async fn index(
...
+    check_user_in_organization(
+        &app_state,
+        authenticated_user,
+        organization_id,
+        request,
+        next,
+    )
+    .await
+
+    // TODO: Check that the authenticated user is a coach
```
jhodapp, line 49: Did you mean to leave this as a TODO? Is it not straightforward to implement this check at this time?

jhodapp (commented):
This looks great, thank you Caleb. A couple of questions inline.

entity_api/src/coaching_relationship.rs
```diff
@@ -15,7 +15,7 @@ use serde::ser::{Serialize, SerializeStruct, Serializer};
 use slugify::slugify;
 
 pub async fn create(
-    db: &DatabaseConnection,
+    db: &impl ConnectionTrait,
```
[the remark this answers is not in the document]
calebbourg: Yeah good thought. I think it probably makes sense to use the `impl` form in other places just for the sake of flexibility. I was moving quickly and didn't want to expand the scope too much for this PR. That's the only real reason I didn't change everything else. I don't think it matters _too_ much either way at the moment except for these cases where we need it. 

Open to what you think.

web/src/protect/organizations/users.rs
calebbourg: I left it yeah a note that we should check eventually. It's not immediately clear to me whether or not this endpoint will be only for coaches in the future until we work out some of the specifics around the user creation workflow. Ex create user -> assign user to coaching relationship later. 
That's why I left it out at the moment. It may be that we decide on separate endpoints for different "flavors" of user creation.

entity_api/src/coaching_relationship.rs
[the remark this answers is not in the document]
jhodapp: Yeah agreed, I don't think we should change it everywhere in this PR. I wanted to first make sure I understand why it's necessary here, and I believe I do. They don't seem to discuss the use of the ConnectionTrait in the SeaORM documentation which is curious. But in the crate API docs, it's definitely there as a trait that can be used.

web/src/protect/organizations/users.rs
jhodapp: Sounds good and reasonable. Would you mind adding your comment to your TODO as helpful context? I'd be happy with that then.

calebbourg: Absolutely! Added d0c6dfd71e1f0352288265bd8555fc54c18e3b30

jhodapp (approved):

PR #115, opened 2025-04-04
Delete coaching sessions

calebbourg:
## Description

Adds endpoint for deleting existing coaching sessions.

The following applies:
- Only users that are coaches for the coaching session are authorized to delete it
- We attempt to delete the associated Tiptap document. If that fails we do not delete the coaching session

### Changes
* Adds `DELETE /coaching_sessions/{coaching_session_id}
* Refactors `domain::coaching_sessions`

### Testing Strategy
 Tested locally against https://github.com/refactor-group/refactor-platform-fe/pull/100

### DIFF
--- a/domain/src/gateway/tiptap.rs
+++ b/domain/src/gateway/tiptap.rs
@@ -1,5 +1,6 @@
-use crate::error::{DomainErrorKind, Error, InternalErrorKind};
+use crate::error::{DomainErrorKind, Error, ExternalErrorKind, InternalErrorKind};
 use log::*;
+use serde_json::json;
 use service::config::Config;
 
 /// HTTP client for making requests to Tiptap. This client is configured with the necessary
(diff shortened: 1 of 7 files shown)


domain/src/gateway/tiptap.rs
```diff
@@ -33,3 +34,81 @@ async fn build_auth_headers(config: &Config) -> Result<reqwest::header::HeaderMa
...
+            Error {
+                source: None,
+                error_kind: DomainErrorKind::Internal(InternalErrorKind::Other),
+            }
+        })?;
+        Ok(Self { client, base_url })
+    }
+
+    pub async fn create(&self, document_name: &str) -> Result<(), Error> {
+        let url = format!(
```
jhodapp, line 57: I suggest pulling this up to the top of this module since it's repeated for `create` and `delete`, which will also help centralize it in case TipTap ever changes their cloud API paths.

jhodapp (changes requested):
Looks fantastic! One minor suggested change to make.

domain/src/gateway/tiptap.rs
calebbourg: Great idea 3e0a566cda9b49ef57e216f199d3b6faa75d8020

jhodapp (approved):

PR #117, opened 2025-04-09
Update users

calebbourg:
## Description

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns

### DIFF
--- a/domain/src/user.rs
+++ b/domain/src/user.rs
@@ -5,8 +5,8 @@ use crate::{
 };
 use chrono::Utc;
 pub use entity_api::user::{
-    create, create_by_organization, find_by_email, find_by_id, find_by_organization, AuthSession,
-    Backend, Credentials,
+    create, create_by_organization, find_by_email, find_by_id, find_by_organization,
+    verify_password, AuthSession, Backend, Credentials,
 };
 use entity_api::{
     coaching_relationship, mutate, organizations_user, query, query::IntoQueryFilterMap, user,
@@ -30,13 +30,15 @@ pub async fn update(
     params: impl mutate::IntoUpdateMap,
 ) -> Result<users::Model, Error> {
     let existing_user = find_by_id(db, user_id).await?;
+
+    let mut params = params.into_update_map();
+    // remove the password and verify it so that we will not attempt to update the password
+    // downstream.
+    let password_to_verify = extract_password(&mut params).await?;
+    verify_password(&password_to_verify, &existing_user.password).await?;
+
     let active_model = existing_user.into_active_model();
-    Ok(mutate::update::<users::ActiveModel, users::Column>(
-        db,
-        active_model,
-        params.into_update_map(),
-    )
-    .await?)
+    Ok(mutate::update::<users::ActiveModel, users::Column>(db, active_model, params).await?)
 }
 
 // This function is intended to be a temporary solution until we finalize our user experience strategy for assigning a new user
@@ -96,4 +98,14 @@ pub async fn delete(db: &DatabaseConnection, user_id: Id) -> Result<(), Error> {
     })?;
 
     Ok(())
+}
+
+async fn extract_password(params: &mut mutate::UpdateMap) -> Result<String, Error> {
+    params
+        .remove("password")
+        .map(|v| v.to_string())
+        .ok_or(Error {
+            source: None,
+            error_kind: DomainErrorKind::Internal(InternalErrorKind::Other),
+        })
 }
--- a/entity_api/src/mutate.rs
+++ b/entity_api/src/mutate.rs
@@ -67,6 +67,13 @@ impl UpdateMap {
         self.map.get(key).and_then(|opt| opt.as_ref())
     }
 
+    /// Removes a key-value pair from the map.
+    ///
+    /// Returns the removed value if it exists, or None if the key is not found.
+    pub fn remove(&mut self, key: &str) -> Option<Value> {
+        self.map.remove(key).and_then(|opt| opt)
+    }
+
     /// Inserts a key-value pair into the map.
     ///
     /// If the key already exists, the value will be overwritten.
--- a/web/src/params/user.rs
+++ b/web/src/params/user.rs
@@ -9,8 +9,10 @@ pub struct UpdateParams {
     pub last_name: Option<String>,
     pub display_name: Option<String>,
     pub github_profile_url: Option<String>,
+    // Note that this will be ignored and is only used here for an extra verification
+    // step
+    pub password: Option<String>,
 }
-
 impl IntoUpdateMap for UpdateParams {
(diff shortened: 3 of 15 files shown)


entity_api/src/mutate.rs
```diff
@@ -59,19 +60,56 @@ impl UpdateMap {
...
     }
 
-    /// Inserts a key-value pair into the map.
-    ///
-    /// If the key already exists, the value will be overwritten.
-    pub fn insert(&mut self, key: String, value: Option<Value>) {
-        self.map.insert(key, value);
+    /// Removes a value from the update map and returns it.
+    /// Returns Error if the key doesn't exist or the value is not a valid string.
+    pub fn remove(&mut self, key: &str) -> Result<String, Error> {
```
calebbourg, line 80: Somewhere down the line I'd like to think through this specific data structure a bit more. It's somewhat unwieldy. It may make sense to just use a `HashMap` directly.. not sure.

web/src/params/user.rs
```diff
@@ -47,3 +46,29 @@ impl IntoUpdateMap for UpdateParams {
         update_map
     }
 }
+
+#[derive(Debug, Deserialize, IntoParams, ToSchema)]
+pub struct UpdatePasswordParams {
```
jhodapp, line 51: I love how closely the frontend and backend code follow each other here on types. This makes code comprehension so much easier.

web/src/protect/users/passwords.rs
```diff
@@ -0,0 +1,30 @@
...
+    Path(user_id): Path<Id>,
+    request: Request,
+    next: Next,
+) -> impl IntoResponse {
+    // check that we are only allowing authenticated users to update their own password
+    if authenticated_user.id == user_id {
+        next.run(request).await
+    } else {
+        error!(
+            "Unauthorized: user_id {} does not match authenticated_user_id {}",
```
jhodapp, line 25:
```suggestion
            "Unauthorized: user_id {} does not match authenticated_user_id {} when attempting to update password",
```

Just a suggestion that may help us observe this security-sensitive code more closely when in production. I'm a little nervous about our login handling since we wrote it ourselves.

web/src/router.rs
```diff
@@ -344,25 +346,49 @@ pub fn overarching_goal_routes(app_state: AppState) -> Router {
...
-            delete(user_session_controller::delete),
+            "/users/:id/password",
+            put(user::password_controller::update_password),
         )
-        .route_layer(login_required!(Backend, login_url = "/login"))
+        .route_layer(from_fn_with_state(
+            app_state.clone(),
+            protect::users::passwords::update_password,
+        ))
+        .route_layer(login_required!(Backend, login_url = "/"))
```
jhodapp, line 384: Great catch!

jhodapp (changes requested):
Just a few security-related suggestions to think about. Let me know if you want to discuss any of them.

domain/src/user.rs
```diff
@@ -39,6 +40,41 @@ pub async fn update(
     .await?)
 }
 
+pub async fn update_password(
```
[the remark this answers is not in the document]
calebbourg: This is all really great and many things I didn't think about at all. I'll go through 1 by 1

- **Error Handling:** Completely agree that the `Other` use here does not give a lot of information. I've intentionally deferred adding more descriptive variants to these errors as I would like to take a step back at some point and take inventory of all of the uses `Other` and see what other appropriate categories of errors there are. I wanted to avoid adding ad-hoc variants "kitchen sink" style. I'd like to address that as a separate PR and issue. 

- **Verification of current_password:**  I double checked this and our use of `password_auth` ensures OWASP recommended cryptography as stated in their [docs](https://docs.rs/password-auth/latest/password_auth/#usage):

>Behind the scenes the crate uses the multi-algorithm support in the [password-hash](https://docs.rs/password-hash/latest/password_hash/) crate to support multiple password hashing algorithms simultaneously. By default it supports Argon2 (using the latest OWASP recommended parameters 8), but it can also optionally support PBKDF2 and scrypt by enabling crate features.

- **Handling of Removed Fields** and **Error Handling for Field Removal**:  This is handled both upstream and downstream. Our [params check ](https://github.com/refactor-group/refactor-platform-rs/pull/117/files#diff-392eeea1dac4f4f383f57f54537630922006dafd94a983768be2c6d19ccf4bdbR52-R54) will fail prior to this step if these fields are not present. If this function returns an error, that will be transformed downstream to an error response to the client. This, as far as I have thought through it, should not panic.

- **Password Hashing**: This is handled by [password_auth](https://docs.rs/password-auth/latest/password_auth)

- **Insertion of Hashed Password into params**: Excellent point. We store this in the data structure as that is the contract between the `domain` layer and the `entity_api` layer. I think I would prefer some sort of global filtering of passwords from logs over handling it this way. Especially if this approach would require breaking these abstraction boundaries. 

- **Lack of Rate Limiting** and **Audit Logging**: Love these. I think we should think through how we might implement in a way that's usable across the web interface.

- **Validation of New Password**: Let's discuss what our requirements would be here.

- **Potential for Overwriting Fields:** This is worth a larger discussion as well as the same issue exists across the entire web interface. We should look into how seaORM or Axum recommends sanitization.

web/src/protect/users/passwords.rs
calebbourg: Updated! fd1af337616628751235f836a2f178f8dfc63d9d

domain/src/user.rs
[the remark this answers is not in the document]
calebbourg: I added an issue to add error variants https://github.com/refactor-group/refactor-platform-rs/issues/120

entity_api/src/mutate.rs
calebbourg: I think it's less the data structure itself and more that the way it's implemented now, we need to explicitly wrap values in `Value` structs which exposes some seaORM constructs that I would prefer be hidden from this layer. We achieve semi decoupling by re-exporting the concepts through the app layers but maybe there's a better way. Not sure.

domain/src/user.rs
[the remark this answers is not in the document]
jhodapp: @calebbourg Would you also mind adding another new issue under the heading "Future robust password handling" and tag as `security` and `enhancement`? And then copy your comment here exactly as is as the issue body? Thank you for responding to each one-by-one.

entity_api/src/mutate.rs
jhodapp: I see a lot of Rust crates reexport from underlying crate types/methods, but I agree the strongest interfaces are those that normally try to avoid this. I wouldn't call this a high priority though as long as it's not feeling overly broken and ugly to work with today.

jhodapp (approved):

PR #129, opened 2025-05-08
Use sea orm migrations

calebbourg:
## Description

This PR aims to accomplish the goal of providing a pattern and path forward for using seaORM's database migration system in all environments.

### Changes
* Adds a `migrations` step to our `docker-compose` setup which runs migrations using seaORM
* Adds a README describing migration process and background context
* Adds a migration to add an initial admin user in non-prod environments

### Testing Strategy
Tested locally (will test in staging as well)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,10 +32,16 @@ RUN apt-get update && apt-get install -y bash && rm -rf /var/lib/apt/lists/*
 RUN useradd -m -s /bin/bash appuser
 WORKDIR /app
 
-# Copy only the necessary release binaries
+# Copy the necessary release binaries
 COPY --from=builder /usr/src/app/target/release/refactor_platform_rs .
-COPY --from=builder /usr/src/app/target/release/migration .
+# Naming this to avoid collision with the migration directory added below
+COPY --from=builder /usr/src/app/target/release/migration migration-binary
 COPY --from=builder /usr/src/app/target/release/seed_db .
+
+# In order to run our initial migration which applies a SQL file directly, we need to
+# make sure the directory exists on the container and copy the SQL file into it.
+RUN mkdir -p /app/migration/src
+COPY --from=builder /usr/src/app/migration/src/refactor_platform_rs.sql /app/migration/src/
 
 # Copy entrypoint script and make it executable
 COPY entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -32,11 +32,17 @@ RUN apt-get update && apt-get install -y bash && rm -rf /var/lib/apt/lists/*
 RUN useradd -m -s /bin/bash appuser
 WORKDIR /app
 
-# Copy only the necessary release binaries
+# Copy the necessary release binaries
 COPY --from=builder /usr/src/app/target/release/refactor_platform_rs .
-COPY --from=builder /usr/src/app/target/release/migration .
+# Naming this to avoid collision with the migration directory added below
+COPY --from=builder /usr/src/app/target/release/migration migration-binary
```
calebbourg, line 38: @lmcdonough @jhodapp note that this probably has implications for how we run migrations in production `migration-binary` vs `migration`

docker-compose.yaml
```diff
@@ -21,6 +19,26 @@ services:
...
+      POSTGRES_HOST: postgres
+      POSTGRES_PORT: ${POSTGRES_PORT}
+      DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:${POSTGRES_PORT}/${POSTGRES_DB}
+      BACKEND_LOG_FILTER_LEVEL: ${BACKEND_LOG_FILTER_LEVEL}
+    depends_on:
+      postgres:
+        condition: service_healthy
+    networks:
+      - backend_network
+    command: ["/app/migration-binary", "up", "-s", "${POSTGRES_SCHEMA}"]
```
calebbourg, line 40: using `migration-binary` here to execute the compiled binary

scripts/rebuild_db.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Define default variables
-DB_NAME=${1:-"refactor_platform"}
+DB_NAME=${1:-"refactor"}
```
calebbourg, line 4: updated for consistency across the platform

Dockerfile
jhodapp:
```suggestion
COPY --from=builder /usr/src/app/target/release/migration migrationctl
```

calebbourg: Updated in e7233f073cccd932925bcc1cc87c07f845b0decf

docker-compose.yaml
calebbourg: updated to use `migrationctl` in e7233f073cccd932925bcc1cc87c07f845b0decf

PR #132, opened 2025-05-20
Small improvements

calebbourg:
## Description
This PR adds some small improvements that have been in the back of my mind and which I've deferred in favor of higher priority work over the last few months. There will likely be more of these PRs.

### Changes
* Adds validation that a Coaching Relationship needs to be unique in the scope of it's organization
* Adds some richness to our Error mechanisms

### Testing Strategy
Tested locally

### DIFF
--- a/domain/src/jwt/mod.rs
+++ b/domain/src/jwt/mod.rs
@@ -53,7 +53,9 @@ pub async fn generate_collab_token(
         );
         Error {
             source: None,
-            error_kind: DomainErrorKind::Internal(InternalErrorKind::Other),
+            error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(
+                "Failed to get collab document name from coaching session".to_string(),
+            )),
         }
     })?;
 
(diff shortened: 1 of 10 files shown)


domain/src/jwt/mod.rs
```diff
@@ -69,15 +71,15 @@ pub async fn generate_collab_token(
         Error {
             source: None,
             // TODO make this InternalErrorKind::ConfigError
```
jhodapp, line 73:
```suggestion
```

jhodapp (changes requested):
Just a tiny change to make with a stale comment. This looks great, thanks for these improvement Caleb.

domain/src/jwt/mod.rs
calebbourg: Good call. Updated in dc80f8bc14f61044a4246d9a4925302489f05bb4 !

jhodapp (approved):

PR #137, opened 2025-05-30
Use vars where appropriate for gh workflows

jhodapp:
## Description
This PR converts an appropriate set of environment secrets into variables. And it also makes the deploy_to_do.yml use the production environment for secrets and variables on GH.

### Changes
* Still allow for local builds to work by setting a *_BUILD_CONTEXT env var for frontend or backend. Also pass RUST_ENV environment string to backend
* Significantly improve the robustness of entrypoint.sh
* Expect RUST_ENV environment string to be set for migrator and refactor_platform_rs binaries. Also don't expose DATABASE_URL on command line, pass it by env var only
* Convert all other appropriate secrets to variables

### Testing Strategy
1. Try to build and deploy using GH Actions

### Concerns
None

### DIFF
--- a/service/src/config.rs
+++ b/service/src/config.rs
@@ -157,15 +157,15 @@ impl Config {
             .expect("No API version string provided")
     }
 
-    pub fn set_database_uri(mut self, database_uri: String) -> Self {
-        self.database_url = Some(database_uri);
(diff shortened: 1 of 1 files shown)


qafui (approved):
Nice work!

PR #146, opened 2025-06-06
Move to a multi-layered docker-compose.yaml setup

jhodapp:
## Description
This PR seeks to take the first steps towards having different docker-compose.yaml service setups between development + staging vs production.

It purposefully make development and staging be the same so that they reflect each other as much as possible, and the main difference from production being just the lack of the postgres service container.

### Changes
* Break out docker-compose.yaml into 2 separate docker-compose files

### Testing Strategy
1. Ensure production build and deployment still works. Verify that only the refactor-platform-frontend and refactor-platform-backend containers start, and the postgres container no longer is present (run `docker ps` on production server to verify)
2. Ensure migrations in production, and development + staging still work
3. Make sure you can run a full development/staging containerized system locally by trying to start it up with the command: `docker compose --env-file .env.local -f docker-compose.yaml -f docker-compose.dev-staging.yaml up` (run `docker ps` locally to verify)

### Concerns
None


docker-compose.dev-staging.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - postgres_data:/var/lib/postgresql/data  # persist data
+    networks:
+      - backend_network
+    healthcheck:                            # wait until DB is ready
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  migrator:
```
calebbourg, line 27: Without knowing much about how docker-compose handles these, do we need:
```yaml
    restart: "no"                           # run once, then exit
    command: []  
```
and any others here?

calebbourg (commented):
@jhodapp one small question

docker-compose.dev-staging.yaml
jhodapp: Great question, I'll double check the migrations by starting fresh on our production DB.

jhodapp: @calebbourg it does not seem necessary. The migrator service came up, did it's thing, and spun down when migrationctl finished.

PR #153, opened 2025-06-10
Add roles

calebbourg:
## Description

This PR adds the concept of a **Role** for users. It defines a default role of `user` along with an `admin` role that we can use to provide elevated permissions.

The role concept should not be conflated with Coach vs Coachee relationships as the latter describes relationships between users while the former describes relationships between a user and that user's organization. We need to be careful in how the two concepts evolve overtime.

**Another thing to note:** This PR updates the migration adding the initial admin user.  The migration began failing when I updated the `entity::users::Model` struct to require a `role` attribute. The addition of the attribute caused the compiler to fail due to the migration not including `role` in it's struct definition. **This is an important consideration when seeding data using a strongly typed language**. As we move into production we want to strictly enforce a policy of not updating historical migrations as they, presumably, have been run in _some_ environment (most importantly production). I updated our historical migration in this PR to insert the seeded records using raw SQL as it allows us to circumvent the type checks which were causing the migrations to fail. Please note that I also added a migration which runs _after_ adding the concept of role and which updates the Admin user to promote their role to `admin`.

In the future we should:
- Always provide some default value for columns we consider required
- Use raw SQL for seeding data in migrations in order to avoid type issues when we update type requirements after the fact.
- **Avoid seeding data using migrations unless absolutely necessary**. Prefer to build UI for adding data.

### Changes
* Adds `role` concept
* Updates `Admin` user to role `admin`
* Updates migrations

### Testing Strategy
Tested Locally

### DIFF
--- a/.github/workflows/deploy_to_do.yml
+++ b/.github/workflows/deploy_to_do.yml
@@ -81,7 +81,7 @@ jobs:
           BACKEND_CONTAINER_NAME=${{ vars.BACKEND_CONTAINER_NAME }}
           # We are not building locally so setting to empty string to quiet warning
           BACKEND_BUILD_CONTEXT=""
-          # Backend service port
+          # Backend daemon port used by the Rust backend to listen to connection on
           BACKEND_PORT=${{ vars.BACKEND_PORT }}
           # Network interface for backend to listen for incoming connections on
           BACKEND_INTERFACE=${{ vars.BACKEND_INTERFACE }}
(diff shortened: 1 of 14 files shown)


migration/src/m20250509_164646_add_initial_user.rs
```diff
@@ -29,45 +24,77 @@ impl MigrationTrait for Migration {
...
+        ) VALUES ($1, $2, $3, $4, $5, $6, $7)
+        RETURNING id
+    "#;
+    let user_row = db
+        .query_one(Statement::from_sql_and_values(
+            DbBackend::Postgres,
+            user_sql,
+            vec![
+                Value::String(Some(Box::new("admin@refactorcoach.com".to_owned()))),
+                Value::String(Some(Box::new("admin".to_owned()))),
```
jhodapp, line 48:
```suggestion
                Value::String(Some(Box::new("Admin".to_owned()))),
```

migration/src/m20250509_164646_add_initial_user.rs
```diff
@@ -29,45 +24,77 @@ impl MigrationTrait for Migration {
...
+                Value::String(Some(Box::new("admin".to_owned()))),
```
jhodapp, line 49:
```suggestion
                Value::String(Some(Box::new("Admin".to_owned()))),
```

calebbourg:
@jhodapp thanks for the review! I made the updates you suggested.

jhodapp (approved):

PR #154, opened 2025-06-12
Get https working for production deployment

### LINKED ISSUE #121: [Feature]: Implement TLS for the entire site user flow
### Suggestion / Feature Request

# Description

We are working with sensitive data and user sessions, so this means we need to encrypt all user session traffic.

# Requirements
1. `refactor.engineer` domain is used as the site's production domain (already purchased)
2. Visiting `http://refactor.engineer` always redirects to `https://refactor.engineer`
3. All interactions and data with the platform in production is always encrypted using latest TLS standard

jhodapp:
## Description

Configure HTTPS support for production deployment by adding nginx reverse proxy with SSL termination, updating deployment scripts, and modifying service configurations to work behind a proxy.

GitHub Issue: Closes #121 

### Changes

- Added nginx reverse proxy configuration with SSL/TLS support and Let's Encrypt certificates
- Updated docker-compose.yaml to include nginx service and remove direct port exposure from backend/frontend
- Enhanced deployment workflow with nginx setup, error logging, and improved validation
- Added health check endpoint (/health) to the backend API
- Modified session security settings to use secure cookies in production
- Updated CORS configuration to work with reverse proxy headers
- Added network flow documentation and container setup guidance
- Fixed container naming for consistent service discovery

### Testing Strategy

1. Deploy to production environment (ability to do this with staging is a future enhancement) and verify HTTPS endpoints work correctly via nginx proxy
2. Test SSL certificate validation and confirm secure connections on port 443
3. Verify API routing through /api/ prefix works and strips correctly to backend
4. Confirm frontend routing works for all non-API requests
5. Test health check endpoint accessibility
6. Validate session management and CORS functionality through the proxy

### Concerns
1. Running locally as a fully dockerized system no longer works and is beyond what I'm prepared to do in this PR. It shouldn't be too much work, but it will be very time-consuming to get right, as it should start up without the nginx service and not use the /api virtual backend API path since nginx is needed for this routing to work right. ([I documented this as a new bug](https://github.com/refactor-group/refactor-platform-rs/issues/158))

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -100,6 +100,7 @@ pub async fn init_server(app_state: AppState) -> Result<()> {
             "X-Forwarded-For".parse::<HeaderName>().unwrap(),
             "X-Forwarded-Proto".parse::<HeaderName>().unwrap(),
             "X-Real-IP".parse::<HeaderName>().unwrap(),
+            "X-Request-ID".parse::<HeaderName>().unwrap(),
         ])
         .expose_headers([ApiVersion::field_name().parse::<HeaderName>().unwrap()])
         .allow_private_network(true)
--- a/nginx/conf.d/refactor-platform.conf
+++ b/nginx/conf.d/refactor-platform.conf
@@ -60,6 +60,8 @@ server {
         proxy_set_header X-Forwarded-Proto $scheme;
         proxy_set_header X-Forwarded-Host $host;
         proxy_set_header X-Forwarded-Port $server_port;
+        # Pass a unique identifier or generate one if not provided in a request (see https://http.dev/x-request-id)
+        proxy_set_header X-Request-ID $http_x_request_id$request_id;
         
         # Handle timeouts
         proxy_connect_timeout 60s;
@@ -104,6 +106,8 @@ server {
         proxy_set_header X-Forwarded-Proto $scheme;
         proxy_set_header X-Forwarded-Host $host;
         proxy_set_header X-Forwarded-Port $server_port;
+        # Pass a unique identifier or generate one if not provided in a request (see https://http.dev/x-request-id)
+        proxy_set_header X-Request-ID $http_x_request_id$request_id;
         
         # Handle WebSocket connections (useful for Next.js development)
         proxy_http_version 1.1;


web/src/lib.rs
```diff
@@ -81,12 +89,17 @@ pub async fn init_server(app_state: AppState) -> Result<()> {
...
             Method::PATCH,
         ])
+        // Essential to allow credentials through a reverse proxy like nginx
         .allow_credentials(true)
         // Allow and expose the X-Version header across origins
         .allow_headers([
             ApiVersion::field_name().parse::<HeaderName>().unwrap(),
             AUTHORIZATION,
             CONTENT_TYPE,
+            // Headers that nginx reverse proxy might forward
```
calebbourg, line 99: This is awesome! Would you mind also adding `x-request-id`? 

It's insanely useful for observability 
https://http.dev/x-request-id

calebbourg (changes requested):
Just the one change to add an `x-request-id` header and some other small thoughts but looks good otherwise!

web/src/lib.rs
jhodapp: Fabulous suggestion, I wasn't aware of this header!

calebbourg: Yes definitely

docker-compose.yaml
```diff
@@ -1,7 +1,33 @@
...
+      # Mount your custom nginx configuration
+      - ./nginx/nginx.conf:/etc/nginx/nginx.conf:ro
+      - ./nginx/conf.d:/etc/nginx/conf.d:ro
+      # Mount your letsencrypt SSL certs
+      - /etc/letsencrypt/live/refactor.engineer/:/etc/letsencrypt/live/refactor.engineer/:ro
+      - /etc/letsencrypt/archive/refactor.engineer/:/etc/letsencrypt/archive/refactor.engineer/:ro
+      - ${SSL_DHPARAMS_PATH}:/etc/letsencrypt/ssl-dhparams.pem:ro
+      # Optional: nginx logs for debugging
+      - ./nginx/logs:/var/log/nginx
+    depends_on:
```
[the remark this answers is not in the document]
jhodapp: No need, it already is indirectly since `rust-app` depends on `migrator`.

web/src/lib.rs
calebbourg: @jhodapp I think it might actually make sense for nginx to to generate the values on each request. It's commonly done at the system edge so that every request coming in includes the header and it's value. The BE then reads the value, adds it to every log statement output for the request,  and adds it back as the `x-request-id` in responses to the FE (or to other services if they exist). That way we can trace the whole round trip for each request via logs and we'll be able to tell which logs belong to which request. It becomes necessary when there are hundreds or thousands of parallel requests being served.

jhodapp: With my change, we actually get both. If it's provided, forward it. If it's not, nginx will automatically generate a UUIDv4 unique id.

PR #171, opened 2025-07-11
Fix Database Rebuilds

calebbourg:
## Description

This PR fixes an issue where updates to our `refactor_platform_rs.dbml` file was breaking the migration process when rebuilding a database (local dev).

When running `scripts/rebuild_db.sh`, updates to the `refactor_platform_rs.dbml` were causing subsequent updates to `refactor_platform_rs.sql` and ultimately to the database schema. This was causing changes to be applied in the base migration prior to attempts to make the same changes in later migrations.

We likely still want to be able to document the _current_ state of the database at any given time and so I modified `scripts/rebuild_db.sh` to refer to `base_` files. These files will preserve a snapshot of the database prior to us moving to incremental migrations. 

These files should be considered static now.

I kept `refactor_platform_rs.dbml` at the current state so that we can continue to maintain up to date documentation on the database schema.

**Note:** I ran into a separate issue where the migration to update our admin user to the `admin` role started failing due to Rust's type system complaining due to  `entity::users::Model` having a `timezone` attribute prior to the column existing in the database. 

We have chosen to solve for this type of issue by using raw SQL when seeding records in migrations. There's some explanation in https://github.com/refactor-group/refactor-platform-rs/pull/153.  I overlooked this situation not realizing that the issue would apply to _all_ operations on actual records in migrations.

### Changes
* Adds `base_refactor_platform_rs.dbml` and `base_refactor_platform_rs.sql` along with updates to `scripts/rebuild_db.sh`
* Updates migration to use raw SQL

### Testing Strategy

Run `scripts/rebuild_db.sh` locally

### Concerns
Is there a way we can enforce raw SQL in migrations that operate on records?

### DIFF
--- a/docs/db/base_refactor_platform_rs.dbml
+++ b/docs/db/base_refactor_platform_rs.dbml
@@ -1,0 +1,143 @@
+// Use DBML to define your database structure
+// Docs: https://dbml.dbdiagram.io/docs
+
+// IMPORTANT: DO NOT UPDATE THIS FILE. IT MAINTAINS A STATIC SNAPSHOT OF THE DATABASE PRIOR TO MOVING TO INCREMENTAL seaORM MIGRATIONS
+// AND IS USED TO CREATE THE BASE SCHEMA WHEN REBUILDING THE DATABASE IN NON-PROD ENVIRONEMENTS.ß
+
+Table refactor_platform.organizations {
+  id uuid [primary key, unique, not null, default: `gen_random_uuid()`]
+  name varchar [not null, note: 'The name of the organization that the coach <--> coachee belong to']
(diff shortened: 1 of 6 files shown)


docs/db/base_refactor_platform_rs.dbml
jhodapp, line 1: @calebbourg would the idea then be to keep this locked in as well? When we make a change to the total DB schema and want to reflect it in the dbml and the png diagram that we export, what would the process be?

jhodapp: Oh I see, yes that makes a lot of sense. We're migration-based through and through now, not relying on the .sql file from the dbml at all anymore.

PR #172, opened 2025-07-11
Add mailersend

calebbourg:
## Description

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns


domain/src/user.rs
```diff
@@ -150,3 +153,75 @@ pub async fn delete(db: &DatabaseConnection, user_id: Id) -> Result<(), Error> {
...
+/// Send a welcome email to a newly created user
+async fn send_welcome_email(config: &Config, user: &users::Model) {
+    log::info!(
+        "Initiating welcome email for user: {} ({})",
+        user.email,
+        user.id
+    );
+
+    // Create email request and send it
+    let result = async {
```
calebbourg, line 182: remove async here?

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+
+/// Response from MailerSend API
+#[derive(Debug, Deserialize)]
+#[allow(dead_code)]
+pub struct SendEmailResponse {
+    pub message_id: Option<String>,
+}
+
+impl SendEmailRequest {
+    pub async fn new(
```
jhodapp, line 45: Does this method need to be `async`? It doesn't appear so either for typing reasons or because it's doing anything asynchronous.

Also it seems it's best practice not to make constructor functions be async.

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        // Get the first recipient's email for personalization
+        let email = to
+            .first()
+            .ok_or_else(|| Error {
+                source: None,
+                error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(
+                    "At least one recipient is required".to_string(),
+                )),
+            })?
```
jhodapp, line 68: Maybe it'd make sense to implement this suggestion, since Claude keeps implementing these more verbose Error construction lines?

  **Repeated Pattern:**
```rust
  // Lines 64-67, 88-93, 167-170, 186-189
  Error {
      source: None,
      error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(
          "Some error message".to_string(),
      )),
  }
```

  **Idiomatic Fix:**
```rust
  // Add error construction helpers
  impl Error {
      pub fn internal(msg: impl Into<String>) -> Self {
          Self {
              source: None,
              error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(msg.into())),
          }
      }

      pub fn config(msg: &str) -> Self {
          Self {
              source: None,
              error_kind: DomainErrorKind::Internal(InternalErrorKind::Config),
          }
      }
  }

  // Usage:
  Error::internal("At least one recipient is required")
```

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+            .email
+            .clone();
+
+        Ok(SendEmailRequest {
+            from,
+            to: to.clone(),
```
jhodapp, line 74: Unnecessary clone - `to` is already owned.

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+                    .get("x-message-id")
+                    .and_then(|v| v.to_str().ok())
+                    .map(|s| s.to_string());
+
+                info!("Email sent successfully to {to_emails:?}, message_id: {message_id:?}");
+            } else {
+                let error_text = response.text().await.unwrap_or_default();
+                warn!("Failed to send email to {to_emails:?}: {status} - {error_text}");
+            }
+        });
```
jhodapp, line 149: **Fire-and-Forget Pattern Concerns**

  Current (mailersend.rs:113-150):
```rust
  pub async fn send_email(&self, request: SendEmailRequest) {
      tokio::spawn(async move {
          // No way to track or handle errors
      });
  }
```

  **More Idiomatic Approach:**
```rust
  pub fn send_email(&self, request: SendEmailRequest) -> JoinHandle<Result<(), Error>> {
      let client = self.client.clone();
      let base_url = self.base_url.clone();

      tokio::spawn(async move {
          // Return Result for caller to optionally handle
          let response = client.post(&url).json(&request).send().await
              .map_err(|e| Error::network(e))?;
          // ...
          Ok(())
      })
  }
```

 • This function returns a JoinHandle, which allows the caller to await the task and handle its result.
 • The caller can choose to ‎`.await` the handle or ignore it, but the option is there.
 • This is more idiomatic in Rust async code, as it gives control to the caller and makes error handling possible.

domain/src/gateway/mailersend.rs
jhodapp: Another suggestion is to change this to use a Builder pattern, e.g.:

  **Current: Complex constructor with many parameters**
```rust
  SendEmailRequest::new(from, to, subject, template_id, personalization_data)
```

  **Idiomatic Builder Pattern:**
```rust
  #[derive(Default)]
  pub struct SendEmailRequestBuilder {
      from: Option<EmailRecipient>,
      to: Vec<EmailRecipient>,
      subject: Option<String>,
      template_id: Option<String>,
      personalization: HashMap<String, String>,
  }

  impl SendEmailRequestBuilder {
      pub fn from(mut self, email: impl Into<String>) -> Self {
          self.from = Some(EmailRecipient {
              email: email.into(),
              name: None,
          });
          self
      }

      pub fn add_recipient(mut self, recipient: EmailRecipient) -> Self {
          self.to.push(recipient);
          self
      }

      pub fn build(self) -> Result<SendEmailRequest, Error> {
          // Validation here
      }
  }
```

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+pub struct SendEmailResponse {
+    pub message_id: Option<String>,
+}
+
+impl SendEmailRequest {
+    pub async fn new(
+        from: EmailRecipient,
+        to: Vec<EmailRecipient>,
+        subject: String,
+        template_id: String,
```
jhodapp, line 49: It's probably worth following this advice, but I'll leave it to you to make the call on this suggestion from Claude:

**Missing Type Safety**

  **Current: Using raw strings for template IDs**
```rust
  template_id: String,
```

  **Type-Safe Approach:**
```rust
  #[derive(Debug, Clone)]
  pub struct TemplateId(String);

  impl TemplateId {
      pub fn new(id: impl Into<String>) -> Result<Self, Error> {
          let id = id.into();
          if id.is_empty() {
              return Err(Error::validation("Template ID cannot be empty"));
          }
          Ok(Self(id))
      }
  }
```

domain/src/user.rs
```diff
@@ -150,3 +153,64 @@ pub async fn delete(db: &DatabaseConnection, user_id: Id) -> Result<(), Error> {
...
+    Ok(new_user)
+}
+
+/// Send a welcome email to a newly created user
+async fn send_welcome_email(config: &Config, user: &users::Model) -> Result<(), Error> {
+    log::info!(
+        "Initiating welcome email for user: {} ({})",
+        user.email,
+        user.id
+    );
```
jhodapp, line 179: **API Ergonomics Issues**

  **Current (user.rs:157-171):**
```rust
  pub async fn create_by_organization(
      db: &DatabaseConnection,
      config: &Config,  // Why pass config here?
      organization_id: Id,
      user_model: users::Model,
  ) -> Result<users::Model, Error>
```

  **Better Separation of Concerns:**
```rust
  // Keep user creation pure
  pub async fn create_by_organization(
      db: &DatabaseConnection,
      organization_id: Id,
      user_model: users::Model,
  ) -> Result<users::Model, Error>

  // Separate email concern
  pub async fn send_welcome_email_for_user(
      config: &Config,
      user: &users::Model,
  ) -> Result<(), Error>
```

domain/src/user.rs
```diff
@@ -150,3 +153,64 @@ pub async fn delete(db: &DatabaseConnection, user_id: Id) -> Result<(), Error> {
...
+
+    let mailersend_client = MailerSendClient::new(config).await?;
```
jhodapp, line 181: Something to consider. Obviously not a problem today, but could become a fairly big memory hog when it doesn't need to be, especially since you're spawning sub-threads in the mail sender anyway.

  **Current: Creating new client for each email**
```rust
  let mailersend_client = MailerSendClient::new(config).await?;
```

  **Better: Client should be created once and reused**
```rust
  // In AppState or similar
  pub struct EmailService {
      client: Arc<MailerSendClient>,
  }
```

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+use std::collections::HashMap;
+
+/// MailerSend API client for sending transactional emails
+pub struct MailerSendClient {
+    client: reqwest::Client,
+    base_url: String,
+}
+
+/// Email recipient with name and email address
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
jhodapp, line 15: Add `PartialEq`, `Eq`.

domain/src/gateway/mailersend.rs
```diff
@@ -0,0 +1,313 @@
...
+            template_id,
+            personalization: vec![Personalization {
+                email,
+                data: personalization_data,
+            }],
+        })
+    }
+
+    /// Validate email address and return error if invalid
+    fn validate_email(email: &str) -> Result<(), Error> {
```
jhodapp, line 85: **Use impl AsRef<str> for Flexibility**

The `impl AsRef<str>` pattern is a Rust idiom that makes functions more flexible by accepting multiple string-like types without requiring callers to convert them.

```rust
    fn validate_email(email: impl AsRef<str>) -> Result<(), Error> {
        if !EmailAddress::is_valid(email.as_ref()) {
            // ...
        }
    }
```

domain/src/user.rs
calebbourg: yeah didn't think of this at all. You're much more experienced with this type of thing so, potentially naive question I would have is:
if we consider this the same referenced object in memory, will it continue to spawn new threads every time it's invoked? 
If yes: does it make a difference that the base object in memory is copied or not in that case?
If no: will we bottleneck email dispatches?

Or will `MailerSendClient` be an object in only a single instance of the app in which case emails will be invoked sequentially because the single user using that instance of the app is making sequential actions even though the emails themselves are dispatched asynchronously relative to the calling thread? 
More for my academic understanding than anything

domain/src/gateway/mailersend.rs
calebbourg: Updated to use builder pattern 3183118cd34503bffdac15a9dc3160228247c070

domain/src/gateway/mailersend.rs
calebbourg: Upon further reflection, I think I may hold off on this one since it will bloat this PR quite a bit to update every place we're using the current pattern and I'd like to avoid having multiple patterns for the same mechanism in the codebase if possible.
I do think this is worth implementing, however so I created an issue for it https://github.com/refactor-group/refactor-platform-rs/issues/182

domain/src/gateway/mailersend.rs
calebbourg: Added 3183118cd34503bffdac15a9dc3160228247c070

domain/src/gateway/mailersend.rs
calebbourg: Added 3dfacd1aa7d60b2e7b3e5989cae878a333b1a51c

domain/src/user.rs
```diff
@@ -150,3 +153,59 @@ pub async fn delete(db: &DatabaseConnection, user_id: Id) -> Result<(), Error> {
...
+
+    // Send welcome email in the background
+    send_welcome_email(config, &new_user).await?;
+
+    Ok(new_user)
+}
+
+/// Send a welcome email to a newly created user
+async fn send_welcome_email(config: &Config, user: &users::Model) -> Result<(), Error> {
+    log::info!(
```
jhodapp, line 175: No need to use `log::` for these logging macros, I'm guessing Claude spelled these out explicitly.

jhodapp (changes requested):
Just a few more things to consider to finish out each thread and one minor new thing that I didn't notice last time.

domain/src/gateway/mailersend.rs
calebbourg: I think your thought process makes sense. I updated it in bee8c3e8c6e11f176e6961a3cf3b27180b1d07e8

domain/src/user.rs
calebbourg: If you don't mind, I'd like to think through this one a bit more and we can tackle deciding whether or not to send the welcome email at a separate time.

The main reason is it feels a bit like leaking context boundaries. The way it works right now, sending an email is an implementation detail of the `domain::users` module. The suggestion here requires email dispatch to be exported out of `domain::users` as a public part of the module's API. That's what doesn't feel quite right to me.

calebbourg: Actually, what would you think about just having a `domain::emails` module that we can use to do this sort of thing?

```rust
use domain::{user as UserApi, users, emails as EmailAPI, Id};

let user = UserApi::create_by_organization(
        app_state.db_conn_ref(),
        organization_id,
        user_model,
    )
    EmailsAPO::send_welcom_email(&app_state.config, user
```

jhodapp: I love it, that's a perfect architecture for these two things to keep a clean and logical separation of concerns.

web/src/controller/organization/user_controller.rs
```diff
@@ -67,14 +67,13 @@ pub(crate) async fn create(
...
-        organization_id,
-        user_model,
-    )
-    .await?;
+    let user =
+        UserApi::create_by_organization(app_state.db_conn_ref(), organization_id, user_model)
+            .await?;
     info!("User created: {user:?}");
+
+    EmailsAPI::send_welcome_email(&app_state.config, &user).await?;
```
jhodapp, line 75: I love this so much, the separation of concerns is _very_ clear now.

jhodapp (approved):

PR #176, opened 2025-07-30
Sort and group coaching sessions

jhodapp:
## Description

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns


jhodapp:
Claude's PR review recommendations that are implemented:

1. Add Index Documentation: Document required database indexes for sortable columns
2. Enhance OpenAPI Spec: Add enum value examples for frontend developers
3. Parameter Validation: Prevent partial sorting parameter combinations

web/src/params/coaching_session.rs
```diff
@@ -1,15 +1,64 @@
...
 use domain::Id;
 use domain::{IntoQueryFilterMap, IntoUpdateMap, QueryFilterMap, UpdateMap};
-use sea_orm::Value;
+use sea_orm::{Order, Value};
 use serde::Deserialize;
 use utoipa::{IntoParams, ToSchema};
 
+#[derive(Debug, Deserialize, utoipa::ToSchema)]
+#[schema(example = "date")]
+pub(crate) enum CoachingSessionSortField {
```
calebbourg, line 12: Can we chance this to just `SortField` since the `CoachingSession` part will be encoded in the module name `..params::coaching_session::SortField`

domain/src/coaching_session.rs
```diff
@@ -70,6 +71,29 @@ pub async fn find_by(
...
+    let coaching_sessions = query::find_by_with_sort::<
+        coaching_sessions::Entity,
+        coaching_sessions::Column,
+    >(db, query_filter_map, sort_column, sort_order)
+    .await?;
+
+    Ok(coaching_sessions)
+}
+
+pub trait CoachingSessionSortParams {
```
calebbourg, line 92: This trait (or at least the concept and functionality), I propose, belongs in `entity_api/src/query.rs` and should be made to be generic in the same way that `IntoQueryFilterMap` is used. That way other entities can be made sortable.

entity_api/src/query.rs
```diff
@@ -105,3 +107,32 @@ where
 
     Ok(query.all(db).await?)
 }
+
+/// Find all records of an entity by the given query filter map with optional sorting.
+pub async fn find_by_with_sort<E, C>(
```
calebbourg, line 112: Would it be possible to add sorting to `find_by` above and apply sorting if the proper parameters, etc. are present? Seems like that's actually what this is doing so maybe we can just update `find_by` to look more like this? 

Ideally we would either encode sorting into `IntoQueryFilterMap` and use that trait instead of `CoachingSessionSortParams` above or we make the concept of sorting generic and have it be it's own trait within this module. I don't have a solid implementation in mind but that's the direction I hope we can go. Let me know if I can help at all with thinking through that or if it just straight up doesn't make sense.

migration/src/m20250730_000000_add_coaching_sessions_sorting_indexes.rs
```diff
@@ -0,0 +1,119 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
```
calebbourg, line 7: This looks good at face value. I haven't tested it yet. One thought that comes to mind up front is that this will probably freeze the coaching sessions table while the index is created in order to provide atomicity in case a rollback is necessary.  Other migration frameworks I've used have some mechanism for disabling these kinds of transactions so not sure if seaORM provides that. I think they're called DDL transactions. 

Not necessary for this one since we likely won't have users creating coaching sessions while the migration is running in production. Just something to get on paper for us to think about in the future.

docs/db/sorting_indexes.md
```diff
@@ -0,0 +1,66 @@
+# Database Indexes for Sorting
```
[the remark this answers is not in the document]
jhodapp: Why thank you. :)

web/src/params/coaching_session.rs
```diff
@@ -1,15 +1,64 @@
...
 #[derive(Debug, Deserialize, IntoParams)]
 pub(crate) struct IndexParams {
     pub(crate) coaching_relationship_id: Id,
     pub(crate) from_date: NaiveDate,
     pub(crate) to_date: NaiveDate,
+    pub(crate) sort_by: Option<CoachingSessionSortField>,
+    pub(crate) sort_order: Option<SortOrder>,
+}
+
+impl IndexParams {
```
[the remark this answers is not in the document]
jhodapp: Agreed, and done.

web/src/params/coaching_session.rs
jhodapp: Done.

entity_api/src/query.rs
jhodapp: Done.

migration/src/m20250730_000000_add_coaching_sessions_sorting_indexes.rs
jhodapp: This is what Claude reports on this, placing it here for reference since it's an interesting finding:

  **The Issue**

  When creating indexes on PostgreSQL (which you're using), the default behavior is:
  - Acquires ACCESS EXCLUSIVE lock on the table
  - Blocks all reads and writes during index creation
  - Can take minutes/hours on large tables
  - Causes downtime for the application

  **Solution: Concurrent Index Creation**

  PostgreSQL supports CREATE INDEX CONCURRENTLY which:
  - Allows reads/writes during index creation
  - Takes longer but doesn't block traffic
  - Cannot run inside transactions (important for migrations)

jhodapp: Looks like SeaORM's running migrations in transactions make it not possible to create a non-locking index. I added this note to the two new transaction files included in this PR.

jhodapp:
@calebbourg ready for your re-review. Quite a few changes because of the unification and centralization of sorting and grouping, per your PR feedback.

web/src/params/overarching_goal.rs
```diff
@@ -1,13 +1,28 @@
...
 
-use domain::{IntoQueryFilterMap, QueryFilterMap};
+use super::sort::SortOrder;
+use super::WithSortDefaults;
+use domain::{overarching_goals, Id, IntoQueryFilterMap, QueryFilterMap, QuerySort};
+
+/// Sortable fields for overarching goals
+#[derive(Debug, Deserialize, ToSchema)]
+#[schema(example = "title")]
+pub(crate) enum OverarchingGoalSortField {
```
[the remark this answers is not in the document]
jhodapp: A very good catch for code polish. I fully support it.

web/src/params/coaching_session.rs
```diff
@@ -1,15 +1,64 @@
...
+    Date,
+    #[serde(rename = "created_at")]
+    CreatedAt,
+    #[serde(rename = "updated_at")]
+    UpdatedAt,
+}
+
+#[derive(Debug, Deserialize, utoipa::ToSchema)]
+#[schema(example = "desc")]
+pub(crate) enum SortOrder {
```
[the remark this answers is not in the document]
calebbourg: great

calebbourg (approved):
love it

PR #184, opened 2025-08-15
Update the Session on Reads

calebbourg:
## Description
This PR updates our Session configuration to use `with_always_save(true)`

From the docs

> Configures whether unmodified session should be saved on read or not. When the value is true, the session will be saved even if it was not changed.
>
> This is useful when you want to reset [Session](https://docs.rs/tower-sessions/latest/tower_sessions/struct.Session.html) expiration time on any valid request at the cost of higher [SessionStore](https://docs.rs/tower-sessions/latest/tower_sessions/trait.SessionStore.html) write activity and transmitting set-cookie header with each response.
>
>It makes sense to use this setting with relative session expiration values, such as Expiry::OnInactivity(Duration). This setting will not cause session id to be cycled on save.
>
>The default value is false.

### Changes
* Use `with_always_save(true)`

### Testing Strategy
I tested this locally by changing the expiry to 1 minute and:
- confirmed that I was logged out after 1 minute of inactivity.
- confirmed that I _was not_ logged out after 1 minute while using the application

### Concerns
It's not clear why `with_always_save(true)` works and `save()` didn't. I didn't look too deeply. My guess is `save()` bypasses updating the relevant internals of the session data structure.


jhodapp (approved):
Approved. Feel free to merge and deploy after fixing the `cargo fmt` issue.

PR #188, opened 2025-09-16
185 allow organization specific roles

calebbourg:
## Description
This PR adds a `user_roles` table that we will use to provide more granularity to our permissions scheme.

The `user_roles` table is associated with both the `users` table and, optionally, the `organizations` table. All users will have an associated `user_roles`.

As of this PR the available roles are:
- `User` - default role
- `Admin` - admin within the scope of an organization.
- `SuperAdmin` - global admin.

Users with role `Admin` or `User` will have a `user_roles` record with `role` `Admin` or `User` respectively. the `user_roles` record will be scoped to the organization within which the user has that role.

Users with role `SuperAdmin` with have a `user_roles` record with `role` `SuperAdmin` and will _not_ have an `organization` associated with their `user_roles` record.

## Next up
* Incorporating the new table into the API
* Adding `user_roles` records for existing users
* FE UI

### Changes
* Add `user_roles` table via migration
* Add `super_admin` to enum type `role`
* Add `user_roles` entity

#### rebuilding the db:
```bash
scripts/rebuild_db.sh
```
#### roll back migration
```bash
DATABASE_URL=postgres://refactor:password@localhost:5432/refactor sea-orm-cli migrate down -s refactor_platform
```

#### re-apply migration
```bash
DATABASE_URL=postgres://refactor:password@localhost:5432/refactor sea-orm-cli migrate up -s refactor_platform
```

### Concerns
The migration is written mostly in Raw SQL. I opted for this as seaORM's documentation around migrations is still very lacking in actual usable examples.[ The main drawback here is portability between postgres and other database products should we ever want to switch.](https://www.sea-ql.org/SeaORM/docs/migration/writing-migration/#using-raw-sql) It's highly unlikely that that will ever happen.

>You can write migration files in raw SQL, but then you lost the multi-backend compatibility SeaQuery offers.


entity/src/user_roles.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    #[serde(skip_deserializing)]
+    pub id: Uuid,
+    pub role: Role,
+    pub organization_id: Option<Uuid>,
+    pub user_id: Uuid,
+    pub created_at: DateTimeWithTimeZone,
+    pub updated_at: DateTimeWithTimeZone,
+}
```
jhodapp, line 18: **Issue: Missing Validation:**
- No application-level constraints ensuring SuperAdmin roles have `organization_id = NULL`
- No validation that organization-scoped roles have non-NULL `organization_id`

**Suggestion:**
- Add validation logic in the model or service layer:
```rust
impl Model {
  pub fn validate(&self) -> Result<(), ValidationError> {
      match self.role {
          Role::SuperAdmin if self.organization_id.is_some() => {
              Err(ValidationError::new("SuperAdmin cannot be scoped to organization"))
          }
          Role::User | Role::Admin if self.organization_id.is_none() => {
              Err(ValidationError::new("User/Admin roles must be scoped to organization"))
          }
          _ => Ok(())
      }
  }
}
```

**More Detailed Explanation:**
-- ❌ BAD: SuperAdmin incorrectly scoped to an organization
```sql
INSERT INTO user_roles (user_id, organization_id, role) VALUES
  ('user-123', 'some-org-uuid', 'super_admin');
```

-- ❌ BAD: Admin role with no organization (global admin?)
```sql
INSERT INTO user_roles (user_id, organization_id, role) VALUES
  ('user-456', NULL, 'admin');
```

The database schema allows these invalid combinations because:
- `organization_id` is `Option<Uuid>` (nullable)
- There's no constraint enforcing the business rule

_What the Validation Does_

```rust
impl Model {
    pub fn validate(&self) -> Result<(), ValidationError> {
        match self.role {
            // Enforce: SuperAdmin MUST be global (organization_id = NULL)
            Role::SuperAdmin if self.organization_id.is_some() => {
                Err(ValidationError::new("SuperAdmin cannot be scoped to organization"))
            }
            // Enforce: User/Admin MUST be org-scoped (organization_id != NULL)
            Role::User | Role::Admin if self.organization_id.is_none() => {
                Err(ValidationError::new("User/Admin roles must be scoped to organization"))
            }
            _ => Ok(())
        }
    }
}
```

Perhaps adding some unit tests around this also makes sense?

README.md
```diff
@@ -226,5 +226,5 @@ DATABASE_URL=postgres://refactor:password@localhost:5432/refactor_platform sea-o
 Note that to generate a new Entity using the CLI you must ignore all other tables using the `--ignore-tables` option. You must add the option for _each_ table you are ignoring.
 
 ```bash
- DATABASE_URL=postgres://refactor:password@localhost:5432/refactor_platform sea-orm-cli generate entity  -s refactor_platform -o entity/src -v --with-serde both --serde-skip-deserializing-primary-key --ignore-tables {table to ignore} --ignore-tables {other table to ignore}
+ DATABASE_URL=postgres://refactor:password@localhost:5432/refactor sea-orm-cli generate entity  -s refactor_platform -o entity/src -v --with-serde both --serde-skip-deserializing-primary-key --ignore-tables {table to ignore} --ignore-tables {other table to ignore}
```
jhodapp, line 229: Nice catch!

entity/src/user_roles.rs
```diff
@@ -0,0 +1,52 @@
...
+use sea_orm::entity::prelude::*;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(schema_name = "refactor_platform", table_name = "user_roles")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    #[serde(skip_deserializing)]
+    pub id: Uuid,
+    pub role: Role,
```
jhodapp, line 13: **Issue:**
* Two different role enums exist and are inconsistent

**Security Risk:**
  - SuperAdmin users can't be properly validated by existing UserIsAdmin checks
  - Inconsistent role checking across the application
  - Migration adds super_admin enum value but entity doesn't reflect it

```rust
// entity/src/roles.rs (for user_roles table)
pub enum Role {
    User,
    Admin,
    // SuperAdmin  <- Added in migration but missing from enum!
}
```
```rust
// entity/src/users.rs (for users table)
pub enum Role {
    User,
    Admin,
    // No SuperAdmin option
}
```

**Solution:**
```rust
// entity/src/roles.rs needs updating
#[derive(Debug, Clone, Eq, PartialEq, EnumIter, Deserialize, Default, Serialize, DeriveActiveEnum)]
pub enum Role {
    #[default]
    User,
    Admin,
    #[sea_orm(string_value = "super_admin")]  // ADD THIS
    SuperAdmin,  // ADD THIS
}
```

entity/src/user_roles.rs
jhodapp, line 1: **Issue:**
* Existing UserIsAdmin check only validates Role::Admin, not Role::SuperAdmin.

**Solution:**
* New authorization check or update existing one:

```rust 
pub struct UserIsSuperAdmin;

#[async_trait]
impl Check for UserIsSuperAdmin {
    async fn eval(&self, _app_state: &AppState, authenticated_user: &domain::users::Model, _args: Vec<Id>) -> bool {
        authenticated_user.role == domain::users::Role::SuperAdmin
    }
}
```

```rust
// OR update existing check to handle hierarchies
pub struct UserIsAdminOrSuperAdmin;
```

migration/src/m20250916_060419_add_user_roles_table_and_super_admin.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        manager
+            .get_connection()
+            .execute_unprepared(create_index_sql)
+            .await?;
+
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
```
jhodapp, line 57: Something to consider for the future:

**Issue:** Adding user_roles with CASCADE delete means role history is lost, but this is consistent with existing cleanup behavior for coaching_relationships and organizations_users.

**Recommendation:** Consider audit table instead of preventing this entirely.

entity/src/user_roles.rs
calebbourg: I plan to add validation in the next iteration when I add access to the data via web API

migration/src/m20250916_060419_add_user_roles_table_and_super_admin.rs
```diff
@@ -0,0 +1,74 @@
...
+        )";
+
+        manager
+            .get_connection()
+            .execute_unprepared(create_table_sql)
+            .await?;
+
+        // 3. Create unique index to prevent duplicate role assignments
+        let create_index_sql = "CREATE UNIQUE INDEX IF NOT EXISTS user_roles_user_org_role_unique 
+            ON refactor_platform.user_roles(user_id, organization_id, role)";
```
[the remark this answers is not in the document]
calebbourg: Very interesting. Would not have considered that 28e174cd99593daf9670e4fcabd5d9cca18dbb4b

jhodapp:
> @jhodapp Thank you! I believe I've addressed all of the comments.

Thanks! I'll re-review tomorrow morning.

jhodapp:
> @jhodapp I added the check [here](https://github.com/refactor-group/refactor-platform-rs/pull/188/files#diff-f1c7493b30d00626c9731e449df73adf474beaaa84fcb142cd9928c92b47b570R172-R181)
> 
> There are currently no Super Admin users and so this check is not used at the moment.

Thanks for doing this. I think it's a good thing to add ahead of time.

PR #189, opened 2025-09-19
185 add user role data to web api

calebbourg:
## Description
This PR adds a new `roles` field to the `users` table which is a Vector of `UserRoles`. 

Note: this did not require a migration due to the fact that the `roles` field is computed at runtime as a join between the `users` table and the `user_roles` table.

Many of the changes in this PR are due to running `cargo fmt` after a local update of my Rust version

### Next Up
* Add migration to port current `role` values into `roles` for existing users
* Add migration to remove `role` field from `users` table

### Changes
* Adds `roles` filed to `User` model
* Adds `user_roles` validation
* Updates formatting due to running `cargo fmt` with newer Rust version

### Testing Strategy
Tested locally by successfully adding a new user

### DIFF
--- a/entity_api/src/user.rs
+++ b/entity_api/src/user.rs
@@ -202,7 +202,6 @@ impl AuthnBackend for Backend {
         &self,
         creds: Self::Credentials,
     ) -> Result<Option<Self::User>, Self::Error> {
-
         match find_by_email(self.db.as_ref(), &creds.email).await? {
             Some(user) => authenticate_user(creds, user).await,
             None => Err(Error {


jhodapp (approved):
A few changes here, and I'm going to click approve just so that you're able to merge this and deploy it when you're ready while I'm out of office next week.

Here's a few of other things to consider:
1. Consider adding audit logging for role changes that sticks out visually in the logs
2. Add new integration tests for role validation logic

And lastly, I'm going to paste in here exactly what Claude identified around test mock consistency for you to consider:

**Test Mock Complexity Issue Analysis**

**🔍 The Problem**

The authentication system now has two different query patterns that need different mock responses:

1. Authentication Flow (authenticate() method - line 208)

```rust
// Uses find_by_email() which does find_with_related(user_roles::Entity)
match find_by_email(self.db.as_ref(), &creds.email).await? {
    Some(user) => authenticate_user(creds, user).await,
    //...
}
```

Expected Query: Complex JOIN query returning (User, Vec<UserRole>) tuple
```sql
SELECT "users"."id" AS "A_id", ..., "user_roles"."role" AS "B_role", ...
FROM "users" LEFT JOIN "user_roles" ON "users"."id" = "user_roles"."user_id"
WHERE "users"."email" = $1 ORDER BY "users"."id" ASC
```

2. Session Lookup (get_user() method - line 220)

```rust
// Uses simple Entity::find_by_id().one() - NO role join
let user: Option<Self::User> = Entity::find_by_id(*user_id).one(self.db.as_ref()).await?;
```

Expected Query: Simple query returning just User
```sql
SELECT "users"."id", "users"."email", ... FROM "users" WHERE "users"."id" = $1 LIMIT $2
```

**🚨 Current Test Problem**

Looking at the auth middleware test changes:

```rust
let db = Arc::new(
    MockDatabase::new(DatabaseBackend::Postgres)
        // ✅ CORRECT: Tuple for find_with_related query
        .append_query_results([[(test_user.clone(), None::<user_roles::Model>)]]) // For find_with_related in authentication

        // ⚠️  POTENTIAL ISSUE: Simple user for get_user
        .append_query_results([[test_user.clone()]]) // For get_user after login (simple find)
        .append_query_results([[test_user.clone()]]) // For session user lookup in protected route (simple find)
        .into_connection(),
);
```

**⚠️ Specific Issues to Highlight**

Problem 1: Query Shape Mismatch

The test assumes get_user() uses a simple query, but if the implementation changes to also load roles (which would be consistent), the mock would fail.

Problem 2: Inconsistent Data Loading

- Authentication: User gets loaded WITH roles via find_with_related
- Session lookup: User gets loaded WITHOUT roles via simple find_by_id

This creates inconsistent user objects in the same authentication flow:
// During login: user.roles = [some roles from DB]
// During session lookup: user.roles = [] (empty, not populated)

Problem 3: Brittle Test Design

The test now requires intimate knowledge of:
- Which functions use find_with_related (return tuples)
- Which functions use simple queries (return single entities)
- The exact order of database calls in the authentication flow

**💡 Recommendations**

**1. Fix the Consistency Issue**

```rust
// In entity_api/src/user.rs - make get_user() consistent
async fn get_user(&self, user_id: &UserId<Self>) -> Result<Option<Self::User>, Self::Error> {
    debug!("** get_user(): {:?}", *user_id);

    // ✅ Use find_by_id to load roles consistently
    find_by_id(self.db.as_ref(), *user_id)
        .await
        .map(Some)
        .or_else(|_| Ok(None))
}
```

**2. Improve Test Robustness**

```rust
// Create a test helper for consistent user mocking
fn create_test_user_with_roles() -> (users::Model, Vec<user_roles::Model>) {
    let user = users::Model {
        // ... user fields
        roles: vec![], // Will be populated by find_with_related
    };

    let roles = vec![
        user_roles::Model {
            role: roles::Role::User,
            organization_id: Some(Id::new_v4()),
            user_id: user.id,
            // ... other fields
        }
    ];

    (user, roles)
}

// In tests - use consistent mocking pattern
let (test_user, test_roles) = create_test_user_with_roles();
let db = Arc::new(
    MockDatabase::new(DatabaseBackend::Postgres)
        .append_query_results([[(test_user.clone(), Some(test_roles[0].clone()))]])
        .append_query_results([[(test_user.clone(), Some(test_roles[0].clone()))]])
        .append_query_results([[(test_user.clone(), Some(test_roles[0].clone()))]])
        .into_connection(),
);
```

**3. Alternative: Create Mock Helper Functions**

```rust
// Test helper to abstract away query complexity
impl MockDatabase {
    fn with_user_auth_flow(self, user: users::Model, roles: Vec<user_roles::Model>) -> Self {
        self
            .append_query_results([[(user.clone(), roles.get(0).cloned())]]) // find_with_related
            .append_query_results([[user.clone()]]) // simple find (if still used)
            .append_query_results([[user.clone()]]) // another simple find
    }
}
```

**🎯 Bottom Line**

The test mock complexity issue is about maintaining consistency between:
1. Data loading patterns (with/without roles)
2. Mock expectations (tuples vs single entities)
3. Authentication flow behavior (consistent user data throughout)

The current implementation works but is fragile and inconsistent. The recommendations above would make the system more robust and the tests more maintainable.

PR #200, opened 2025-10-22
Check org id for org admins

calebbourg:
Description

  This PR updates authorization logic to properly check organization-specific admin roles instead of relying on global admin
  status. The UserIsAdmin check now validates that a user has admin privileges within the specific organization being accessed.

  Changes

  - Updated UserIsAdmin check to accept organization_id argument and verify admin role within that specific organization
  - Added support for SuperAdmin role with global access (no organization_id constraint)
  - Modified coaching relationship and user endpoints to pass organization_id to UserIsAdmin predicate
  - Removed user-specific filtering from find_by_organization_with_user_names query - authorization now handles access control
  - Updated function signature to remove unused user_id parameter from coaching relationship queries

  Concerns

  - This changes the behavior of the coaching relationship index endpoint - it will now return all coaching relationships for
  an organization (for admins) rather than filtering to only those where the authenticated user is a coach or coachee


web/src/protect/mod.rs
jhodapp: Relevant but outside of your PR, Claude caught a code doc update to make:

Update documentation examples
```
  - ///     Predicate::new(UserIsAdmin, vec![]),
  + ///     Predicate::new(UserIsAdmin, vec![organization_id]),
```

jhodapp: Add unit tests for `UserIsAdmin::eval()` - Important for security-critical code

jhodapp: Documentation comment at line 39 still shows old single-role pattern - could be updated to reflect new multi-role system

jhodapp (changes requested):
Looking great, just a few minor improvements to consider.

entity_api/src/organization.rs
```diff
@@ -81,19 +79,39 @@ pub async fn find_by(
...
+        .await?
+        .is_some();
+
+    warn!("User {} is_super_admin: {}", user_id, is_super_admin);
+
+    let organizations = if is_super_admin {
+        // Super admins have access to all organizations
+        let orgs = Entity::find().all(db).await?;
+        orgs
+    } else {
```
jhodapp: Good, I like the else case being just a regular user here. Good secure design by default.

entity_api/src/organization.rs
```diff
@@ -64,13 +64,11 @@ pub async fn find_by(
     db: &impl ConnectionTrait,
     params: HashMap<String, String>,
 ) -> Result<Vec<Model>, Error> {
-    let mut query = Entity::find();
-
-    for (key, value) in params {
+    if let Some((key, value)) = params.into_iter().next() {
```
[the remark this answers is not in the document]
calebbourg: The original intention for this function was to allow for other options for filtering. `"name"` for example. We have yet to find the need for it. I think, if it's all the same to you, I'll keep it like this for now mostly just to not add more to the mix in this PR.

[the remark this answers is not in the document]
calebbourg: I did return it to it's original form b8cdc2e333d49d245b3b597933eaed10e201328e as I think it reads better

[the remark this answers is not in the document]
jhodapp: Haha sounds good. :)

PR #207, opened 2025-12-09
Fix relationship data issue

calebbourg:
## Description
  Implement role-based access control for coaching relationships to ensure users can only view relationships they are authorized to see based on
  their role permissions.

  ### Changes
  * Add `find_by_user_and_organization_with_user_names` function in entity_api layer to filter coaching relationships by user
  * Implement `find_by_organization_for_user_with_user_names` in domain layer with role-based filtering logic:
    - SuperAdmins: see all relationships in any organization
    - Org Admins: see all relationships in their organization
    - Regular users: see only relationships where they are coach or coachee
  * Update coaching relationships index endpoint to use the new role-aware function
  * Add `Clone` derive to `CoachingRelationshipWithUserNames` for test extensibility
  * Add mock feature to domain Cargo.toml for future testing support

  ### Testing Strategy

  Manual tested by:
  1. Creating users with different roles (SuperAdmin, Admin, regular user)
  2. Creating coaching relationships between users in the same and different organizations
  3. Querying the coaching relationships endpoint with each user and verifying they only see authorized relationships

### DIFF
--- a/domain/src/coaching_relationship.rs
+++ b/domain/src/coaching_relationship.rs
@@ -2,7 +2,7 @@ use crate::coaching_relationships::Model;
 use crate::error::{DomainErrorKind, EntityErrorKind, Error, InternalErrorKind};
 use entity_api::query::{IntoQueryFilterMap, QuerySort};
 use entity_api::{coaching_relationships, query};
-use sea_orm::DatabaseConnection;
+use sea_orm::{DatabaseConnection, TransactionTrait};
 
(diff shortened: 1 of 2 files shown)


jhodapp (changes requested):
Got a few recommendations for improving the security and performance of this, otherwise looking great. Thanks!

jhodapp (approved):
Looks great, thanks for the fix!

PR #214, opened 2025-12-29
Add action assignees support

jhodapp:
## Summary
Add support for assigning one or more users (coach and/or coachee) to Actions via a new junction table, plus a new endpoint to fetch all actions assigned to a user.

### Database
- Add `action_assignees` junction table migration with foreign keys to actions and users
- Update DBML documentation

### Entity Layer
- Create `action_assignees` entity with SeaORM relations
- Add `ActionWithAssignees` struct combining action data with assignee IDs

### Entity API Layer
- Add CRUD operations for action assignees (`create`, `delete`, `set_assignees`)
- Add `find_by_user_id` and `find_action_ids_by_user_id` for reverse lookups
- Add `find_by_assignee_with_assignees` to fetch actions assigned to a user

### Web Layer
- Update action controller to accept/return `assignee_ids` in requests/responses
- **New endpoint**: `GET /users/{user_id}/assigned-actions` - fetches all actions assigned to a user across all coaching sessions

## New Endpoint

```
GET /users/{user_id}/assigned-actions
```

Returns all actions where the specified user is an assignee (via the junction table), regardless of which coaching session they belong to. Each action includes its full assignee list.

**Response format:**
```json
{
  "status_code": 200,
  "data": [
    {
      "id": "...",
      "coaching_session_id": "...",
      "body": "Complete project proposal",
      "status": "InProgress",
      "due_by": "2024-01-15T00:00:00Z",
      "assignee_ids": ["user-id-1", "user-id-2"]
    }
  ]
}
```

## Test Plan
- [ ] Run `cargo clippy` and `cargo fmt`
- [ ] Run `cargo test -p entity_api --features mock` to verify unit tests pass
- [ ] Deploy and test creating/updating actions with assignees via API
- [ ] Test new `/users/{id}/assigned-actions` endpoint
- [ ] Verify cascade deletes work when action or user is deleted

## Deployment Notes
- Migration must be run before deploying frontend changes that use assignee_ids
- New endpoint is protected by authentication (user can only access their own assigned actions)

### DIFF
--- a/entity_api/src/action.rs
+++ b/entity_api/src/action.rs
@@ -221,6 +221,58 @@ pub async fn find_by_id_with_assignees(
     })
 }
 
+/// Finds all actions assigned to a specific user across all coaching sessions.
+///
+/// This queries the `action_assignees` junction table to find all actions
+/// where the given user is an assignee, then fetches the full action data
+/// with all assignee IDs for each action.
+///
+/// # Arguments
+///
+/// * `db` - Database connection
+/// * `user_id` - The user ID to find assigned actions for
+///
+/// # Returns
+///
+/// A vector of `ActionWithAssignees` containing each action and its assignee IDs.
+///
+/// # Errors
+///
+/// Returns `Error` if the database query fails.
+pub async fn find_by_assignee_with_assignees(
+    db: &DatabaseConnection,
+    user_id: Id,
+) -> Result<Vec<ActionWithAssignees>, Error> {
+    debug!("Finding actions assigned to user_id={user_id}");
+
+    // Get all action IDs where this user is assigned
+    let action_ids = action_assignee::find_action_ids_by_user_id(db, user_id).await?;
+
+    if action_ids.is_empty() {
+        return Ok(vec![]);
+    }
+
+    // Fetch all actions by their IDs
+    let actions = Entity::find()
+        .filter(entity::actions::Column::Id.is_in(action_ids))
+        .all(db)
+        .await?;
+
+    // Build ActionWithAssignees for each action
+    let mut results = Vec::with_capacity(actions.len());
+    for action in actions {
+        let assignee_ids = action_assignee::find_user_ids_by_action_id(db, action.id).await?;
+        results.push(ActionWithAssignees {
+            action,
+            assignee_ids,
+        });
+    }
+
+    debug!("Found {} actions assigned to user {user_id}", results.len());
+
+    Ok(results)
+}
+
 #[cfg(test)]
 // We need to gate seaORM's mock feature behind conditional compilation because
 // the feature removes the Clone trait implementation from seaORM's DatabaseConnection.
--- a/web/src/protect/users/assigned_actions.rs
+++ b/web/src/protect/users/assigned_actions.rs
@@ -1,0 +1,29 @@
+use crate::{extractors::authenticated_user::AuthenticatedUser, AppState};
+use axum::{
+    extract::{Path, Request, State},
+    http::StatusCode,
+    middleware::Next,
+    response::IntoResponse,
+};
+use domain::Id;
+use log::*;
+
+/// Checks that the `user_id` matches the `authenticated_user.id`
+pub(crate) async fn index(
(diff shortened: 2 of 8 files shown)


docs/db/refactor_platform_rs.dbml
```diff
@@ -111,6 +111,19 @@ Table refactor_platform.actions {
   updated_at timestamptz [not null, default: `now()`]
 }
 
+Table refactor_platform.action_assignees {
```
calebbourg: Conventionally, I believe, join tables like this have names that combine both tables their joining. So this would be `actions_users` or `users_actions`.

entity_api/src/action.rs
```diff
@@ -116,6 +131,148 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Model, Error>
...
+    // Create the action first
+    let action = create(db, action_model, user_id).await?;
+
+    // Set assignees if provided
+    let assignee_ids = if let Some(ids) = assignee_ids {
+        action_assignee::set_assignees(db, action.id, ids).await?;
+        action_assignee::find_user_ids_by_action_id(db, action.id).await?
+    } else {
+        vec![]
+    };
```
calebbourg: Do we need this extra step? the user IDs here should be the same ones that are set in the step above unless I'm mistaken?

Also may be able to simplify this

```suggestion
  assignee_ids.map_or( vec![], |id|   action_assignee::set_assignees(db, action.id, ids).await?;)
```

entity_api/src/action.rs
```diff
@@ -116,6 +131,148 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Model, Error>
...
+
+    // Fetch all actions by their IDs
+    let actions = Entity::find()
+        .filter(entity::actions::Column::Id.is_in(action_ids))
+        .all(db)
+        .await?;
+
+    // Build ActionWithAssignees for each action
+    let mut results = Vec::with_capacity(actions.len());
+    for action in actions {
```
calebbourg: Similar N+1 type of thing that maybe we could use joins?

web/src/protect/users/assigned_actions.rs
```diff
@@ -0,0 +1,29 @@
...
+    if authenticated_user.id == user_id {
+        next.run(request).await
+    } else {
+        error!(
+            "Unauthorized: user_id {} does not match authenticated_user_id {}",
+            user_id, authenticated_user.id
+        );
+        (StatusCode::UNAUTHORIZED, "Unauthorized").into_response()
+    }
+}
```
calebbourg: Can we use [UserIsNotSelf](https://github.com/refactor-group/refactor-platform-rs/blob/main/web/src/protect/mod.rs#L186) here?

calebbourg (changes requested):
Few minor things. But looks great otherwise!

entity_api/src/action.rs
```diff
@@ -116,6 +131,148 @@ pub async fn find_by_id(db: &DatabaseConnection, id: Id) -> Result<Model, Error>
...
+    db: &DatabaseConnection,
+    id: Id,
+    model: Model,
+    assignee_ids: Option<Vec<Id>>,
+) -> Result<ActionWithAssignees, Error> {
+    // Update the action
+    let action = update(db, id, model).await?;
+
+    // Update assignees if specified
+    if let Some(ids) = assignee_ids {
```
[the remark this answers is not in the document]
jhodapp: It looks like our existing if let Some/else patterns in this codebase involve either async operations or conditional mutations (side effects), which don't benefit from map_or. So our current patterns are idiomatic for their use cases.

jhodapp:
@calebbourg I applied 6/8 of your suggested changes into my new [replacement PR](https://github.com/refactor-group/refactor-platform-rs/pull/215). Thank you for your excellent suggestions!

You can find all of these changes [here in this commit](https://github.com/refactor-group/refactor-platform-rs/pull/215/changes/32094b8f52378dcb3f0ad813f32d622b71c84db0).

Closing this PR out as it's been replaced by #215.

PR #215, opened 2026-01-03
Add user coaching role endpoints for What's Due dashboard

jhodapp:
## Description
Add new API endpoints to support the "What's Due" dashboard feature. These endpoints enable the frontend to determine a user's coaching roles, retrieve their coaching relationships, and fetch actions with flexible filtering.

### Changes

**Unified Coaching Relationships Endpoint:**
* Add `GET /users/{user_id}/coaching-relationships` - Returns relationships with optional role filtering
  - `role`: `all` (default), `coach`, or `coachee` - filter by user's role in the relationship
  - Returns relationships with user names for both coach and coachee

**Unified Actions Endpoint:**
* Add `GET /users/{user_id}/actions` with flexible query parameters:
  - `scope`: `sessions` (default) or `assigned` - how user relates to actions
  - `coaching_session_id`: filter to specific coaching session
  - `assignee_filter`: `all` (default), `assigned`, or `unassigned`
  - `status`: filter by action status
  - `sort_by`: `due_by` (default), `created_at`, `updated_at`
  - `sort_order`: `asc` (default) or `desc`

**Infrastructure:**
* Add entity_api layer functions with clean separation from web layer
* Add `FindByUserParams` struct for unified action queries (no web annotations)
* Web layer owns HTTP-specific types (`Scope`, `AssigneeFilter`, `RoleFilter` with serde/utoipa)
* Add unit tests for the new unified queries
* Add protect middleware for authorization (users can only query their own data)

### Testing Strategy
1. Start the backend server locally
2. Authenticate as a user who has coaching relationships
3. Test coaching relationships endpoint:
   - Call `GET /users/{user_id}/coaching-relationships` and verify all relationships returned
   - Call `GET /users/{user_id}/coaching-relationships?role=coach` and verify only coach relationships returned
   - Call `GET /users/{user_id}/coaching-relationships?role=coachee` and verify only coachee relationships returned
4. Test unified actions endpoint:
   - Call `GET /users/{user_id}/actions` and verify all actions from user's sessions are returned (default scope)
   - Call `GET /users/{user_id}/actions?scope=assigned` and verify only actions assigned to the user are returned
   - Call `GET /users/{user_id}/actions?assignee_filter=assigned` and verify only actions with assignees are returned
   - Call `GET /users/{user_id}/actions?assignee_filter=unassigned` and verify only actions without assignees are returned
   - Call `GET /users/{user_id}/actions?sort_by=due_by&sort_order=desc` and verify sorting works
5. Verify unauthorized access is rejected when querying another user's data

### Concerns
None - these are read-only endpoints that follow existing patterns in the codebase.

### DIFF
--- a/entity_api/src/coaching_relationship.rs
+++ b/entity_api/src/coaching_relationship.rs
@@ -298,16 +298,15 @@ pub async fn delete_by_user_id(db: &impl ConnectionTrait, user_id: Id) -> Result
     Ok(())
 }
 
-/// Filter for querying coaching relationships by user's role.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
-pub enum RoleFilter {
-    /// Return all relationships where user is coach or coachee (default)
-    #[default]
-    All,
-    /// Return only relationships where user is the coach
-    Coach,
-    /// Return only relationships where user is the coachee
-    Coachee,
+/// Trait for filtering coaching relationships by user's role.
+///
+/// Implement this trait in the web layer to define role-based filtering
+/// while keeping the entity_api layer decoupled from web-specific types.
+pub trait RoleFilterable {
+    /// Returns true if only relationships where the user is a coach should be returned.
+    fn filter_coach_only(&self) -> bool;
+    /// Returns true if only relationships where the user is a coachee should be returned.
+    fn filter_coachee_only(&self) -> bool;
 }
 
 /// Finds coaching relationships for a user with optional role filtering.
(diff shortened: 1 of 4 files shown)


entity_api/src/coaching_relationship.rs
```diff
@@ -312,6 +435,22 @@ impl Serialize for CoachingRelationshipWithUserNames {
     }
 }
 
+/// Summary of a user's coaching roles across all relationships.
+///
+/// Provides boolean flags indicating whether the user is a coach or coachee,
+/// along with counts of each type of relationship.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct RolesSummary {
```
calebbourg: By combining the API endpoints we may not need this. Or perhaps it's added inside the `meta` field in the response?

```json
{
  meta: {
  summary: {
    role_summary: {...}
  }
}
  data: [...]
}
```

In that scenario I chose to have `summary` be an object that can contain arbitrary other summaries. This is just to future proof the `summary` key in case we want to use it for other responses.

Let me know what you think about that

web/src/params/user/action.rs
```diff
@@ -1,16 +1,47 @@
...
+    /// Actions from coaching sessions where user is coach or coachee (default)
+    #[serde(rename = "sessions")]
+    #[default]
+    Sessions,
+}
+
+/// Filter for actions by assignee status.
+#[derive(Debug, Clone, Default, Deserialize, ToSchema)]
+#[schema(example = "all")]
+pub(crate) enum AssigneeFilter {
```
calebbourg: Awesome

web/src/params/user/coaching_relationship.rs
```diff
@@ -0,0 +1,39 @@
+use serde::Deserialize;
+use utoipa::{IntoParams, ToSchema};
+
+use domain::Id;
+
+/// Filter for coaching relationships by user's role.
+#[derive(Debug, Clone, Default, Deserialize, ToSchema)]
+#[schema(example = "all")]
+pub(crate) enum RoleFilter {
```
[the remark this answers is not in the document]
jhodapp: Good catch and an excellent proposed solution. It's much more elegant and avoids duplicating code.

Fixed in cc16258

calebbourg (approved):
Swweeeeet!

I'll leave it up to you if you want to make that last change.

PR #220, opened 2026-02-05
Resolve DB connection pool exhaustion causing 500 errors

jhodapp:
## Description

Backend logs showed repeated `ConnectionAcquire(Timeout)` errors causing 500 Internal Server Errors when loading overarching goals, coaching sessions, and other entities. This correlated with the frontend issue where the session selector dropdown showed error states instead of loading indicators.

**Root cause:** The database connection pool was configured with `max_lifetime(8s)` and `idle_timeout(8s)` — 225x and 75x shorter than sqlx's own defaults, respectively. With 8-second lifetimes, connections were destroyed and recreated constantly. Under moderate concurrent load, connections created around the same time would all expire simultaneously, causing a cascading failure where the pool couldn't create new connections fast enough and the `acquire_timeout(8s)` would fire.

| Setting | sqlx Default | Previous Value |
|---------|-------------|----------------|
| `max_lifetime` | 30 min (1800s) | 8s |
| `idle_timeout` | 10 min (600s) | 8s |

`max_lifetime` exists because PostgreSQL servers accumulate per-session resources (parse trees, query metadata caches, thread-local storage). Periodically retiring connections gives the database a chance to clean up. But 8 seconds doesn't give connections time to be used effectively — it's the equivalent of renting a car, driving one block, returning it, and renting a new one. 30 minutes is sqlx's default, balancing efficient reuse with periodic refresh.

### Changes
* **Connection pool settings**: Increase `max_lifetime` from 8s to 1800s (30min) and `idle_timeout` from 8s to 600s (10min) to match sqlx defaults
* **Configurable pool parameters**: Add 6 Clap fields to `Config` (`DB_MAX_CONNECTIONS`, `DB_MIN_CONNECTIONS`, `DB_CONNECT_TIMEOUT_SECS`, `DB_ACQUIRE_TIMEOUT_SECS`, `DB_IDLE_TIMEOUT_SECS`, `DB_MAX_LIFETIME_SECS`) so pool tuning no longer requires recompilation. All values are logged at INFO level on startup.
* **503 for pool exhaustion**: Add `ServiceUnavailable` variant to `EntityErrorKind` and map `EntityApiErrorKind::SystemError` to 503 Service Unavailable instead of 500, across both the domain→web error chain and the protect middleware — tells frontends "try again" vs "something broke"
* **Fix copy-paste error messages**: Protect middleware in `actions.rs`, `notes.rs`, and `agreements.rs` all logged "Error authorizing overarching goals index" regardless of actual endpoint; each now logs its own name
* **OpenAPI 503 documentation**: Add `(status = 503, description = "Service temporarily unavailable")` to utoipa `responses(...)` annotations on all 47 authenticated endpoints across 17 controller files, so the generated OpenAPI spec communicates this possible response to frontend consumers
* **Improve startup error handling**: Replace `panic!` + `.unwrap_or_default()` in `main.rs` and `.unwrap()` in `seed_db.rs` with proper `error!()` logging and `process::exit(1)`
* **Fix clippy warnings**: Resolve all `uninlined_format_args` warnings across `sse`, `entity`, `entity_api`, `migration`, and `web` crates by inlining variables into format strings
* **Add coaching_relationship_id filter**: Sessions endpoint now supports filtering by `coaching_relationship_id` query parameter
* **Fix broken tests**: Update test calls to `find_by_user_with_includes` to match new function signature

### Testing Strategy
* All existing tests pass (`cargo test` — 0 failures)
* `cargo clippy` passes with zero warnings
* `cargo fmt` produces no changes
* Deploy to staging and monitor logs for `ConnectionAcquire(Timeout)` errors — expect significant reduction or elimination
* Verify pool config values appear in startup logs at INFO level
* Verify the session selector dropdown in the frontend loads without showing false error states
* Verify that when pool exhaustion does occur, the backend returns 503 instead of 500
* Verify pool parameters can be overridden via env vars (e.g. `DB_MAX_CONNECTIONS=50`)

### Concerns
* The `connect_timeout` and `acquire_timeout` default to 8s. These control how long to wait for a new connection or an available pool slot, respectively. 8s is reasonable — if a connection can't be established or acquired in 8s, something is genuinely wrong. These can now be tuned per-environment via env vars if needed.
* The protect middleware error handling still returns 500 for non-`SystemError` failures (e.g., `RecordNotFound` when a coaching session ID doesn't exist). This could arguably be a 404, but is existing behavior and out of scope for this fix.

### DIFF
--- a/.claude/coding-standards.md
+++ b/.claude/coding-standards.md
@@ -122,7 +122,7 @@ pub async fn find_by_id_with_relationship(
 pub use entity_api::some_module::find_by_id_with_relationship;
 ```
 
-4. **Use helper methods on `domain::error::Error`** (e.g., `is_service_unavailable()`) in middleware and handlers rather than matching on deeply nested error kind enums directly. This keeps call sites readable and centralizes the matching logic.
+4. **Use `domain_error_into_response()` in protect middleware** (defined in `web/src/error.rs`) to convert domain errors into HTTP responses. This routes through `web::Error`'s `IntoResponse` impl so that all error-to-status-code mapping stays in one place.
 
 ### Async Patterns
 
--- a/domain/src/error.rs
+++ b/domain/src/error.rs
@@ -53,17 +53,6 @@ pub enum ExternalErrorKind {
     Other(String),
 }
 
-impl Error {
-    /// Returns true if this error represents a service unavailability condition
-    /// (e.g. database connection pool exhaustion).
-    pub fn is_service_unavailable(&self) -> bool {
-        self.error_kind
-            == DomainErrorKind::Internal(InternalErrorKind::Entity(
-                EntityErrorKind::ServiceUnavailable,
-            ))
-    }
-}
-
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "Domain Error: {self:?}")


entity/src/user_roles.rs
```diff
@@ -59,14 +59,12 @@ impl ActiveModelBehavior for ActiveModel {
         match self.role.as_ref() {
             &Role::SuperAdmin if self.organization_id.as_ref().is_some() => {
                 Err(DbErr::Custom(format!(
-                    "[before_save] SuperAdmins should not be associated with an organization, insert: {}",
-                    insert
+                    "[before_save] SuperAdmins should not be associated with an organization, insert: {insert}",
```
jhodapp: FYI these kinds of changes were about addressing all outstanding clippy warnings.

entity_api/src/coaching_session.rs
```diff
@@ -625,7 +632,8 @@ mod tests {
 
         let includes = IncludeOptions::none();
         let results =
-            find_by_user_with_includes(&db, user_id, None, None, None, None, includes).await?;
+            find_by_user_with_includes(&db, user_id, None, None, None, None, None, includes)
```
calebbourg: My comment above (defining a struct for filter related stuff) would make it so that we're not passing a ton of `None`s here.

calebbourg (changes requested):
Some excellent improvements here! Thank you for this. I definitely learned some postgres knowledge from the PR description as well.

My main concerns:
- Preserving context boundaries by keeping `EntityErrorApiKind` hidden from `web`. My rule of thumb has been, wherever possible, a layer should only know about the layer directly underneath it. Ex. `domain` knows about `entity_api` but not `entity`. `web` knows about `domain` but not `entity_api`. 

This hasn't always been easy in practice but I want to be disciplined about it wherever we can be.

- deduplicating the matching in `protect` modules
- removing the clippy allow for `too_many_arguments`

This one, I think, also deserves a bit of thought and design to fit in with other patterns and to potentially be reusable.

entity_api/src/coaching_session.rs
```diff
@@ -218,9 +218,11 @@ impl IncludeOptions {
     }
 }
 /// Find sessions by user with optional date filtering, sorting, and related data includes
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
jhodapp: Fixed in 0492c79.

PR #228, opened 2026-02-19
Email notifications for session scheduling and action assignment

jhodapp:
## Description

Follow-up PR to fix and improve the email notification work that was directly pushed to main in these commits:

* `42042d8` feat: add email notifications for session scheduling and action assignment
* `ad26cee` refactor: rename send_welcome_email to notify_welcome_email
* `ff3b7a5` test: add tests for session scheduled and action assigned emails
* `900eacf` docs: add Mermaid architecture diagram to email notifications plan
* `26c121e` feat: wire action assigned email into update handler and add architecture doc
* `6114290` fix: append ?tab=actions to action assigned email session URL

### Changes

* Resolve clippy `too_many_arguments` by bundling params into `ActionEmailContext` struct
* Only notify newly added assignees on action update (diff before/after) using `HashSet`-based set difference for clarity and efficiency
* Make `send_*` email functions private, expose only `notify_*` public API per two-tier architecture
* Use best-effort pattern for welcome email (log failures, don't block user creation)
* Make MailerSend base URL configurable via `MAILERSEND_BASE_URL` env var and centralize default in Config
* Replace silent URL fallbacks with explicit config errors; extract `SessionUrlBuilder` so URL support is derived from the `EmailNotification` trait rather than a runtime bool
* Wire missing email env vars (`FRONTEND_BASE_URL`, template IDs) through docker-compose and fix `${VAR:-}` → `${VAR-}` defaults so clap `default_value` activates correctly
* Improve email API encapsulation and layer boundaries: eliminate redundant DB re-query in `update_with_assignees`, use `impl Into<String>` consistently in builder API
* Fix doc comments, import ordering, and cargo fmt formatting
* Add coding standards: branching policy and function argument limits

### Testing Strategy

* 31 domain tests pass including 26 email notification tests
* `cargo clippy -p domain -- -D warnings` passes clean
* `cargo fmt --check` passes
* Manually tested end-to-end: created an action with assignees, confirmed Mailersend delivered the email with correct personalization and `?tab=actions` deep link

### Concerns

* The bulk of the email notification implementation was pushed directly to main rather than via PR. This PR contains fixes and improvements on top of that work.
* **Notification preferences** (per-user on/off toggle) are deferred to refactor-group/refactor-platform-fe#301.

### DIFF
--- a/web/src/controller/organization/user_controller.rs
+++ b/web/src/controller/organization/user_controller.rs
@@ -74,10 +74,7 @@ pub(crate) async fn create(
             .await?;
     info!("User created: {user:?}");
 
-    // Best-effort welcome email — log failures, don't block user creation
-    if let Err(e) = EmailsAPI::notify_welcome_email(&app_state.config, &user).await {
-        warn!("Failed to send welcome email to {}: {e:?}", user.email);
-    }
+    EmailsAPI::notify_welcome_email(&app_state.config, &user).await;
 
     Ok(Json(ApiResponse::new(StatusCode::CREATED.into(), user)))
 }
--- a/.claude/coding-standards.md
+++ b/.claude/coding-standards.md
@@ -210,6 +210,30 @@ txn.commit().await?;  // Rolls back automatically if we never reach here
 - **service/**: Orchestration layer, complex operations spanning multiple entities
 - **web/**: HTTP handlers, request/response types, routing
 
+### Thin Controllers
+
+Controllers (web handlers) should be **thin orchestrators**: accept a request, call domain logic, and map the result to an HTTP response. Keep side-effect concerns — logging, best-effort error handling, retries — in the domain layer, not in controllers.
+
+```rust
+// ✅ Good — controller delegates side-effect handling to domain
+pub async fn create(...) -> Result<impl IntoResponse, Error> {
+    let user = UserApi::create_by_organization(db, org_id, model).await?;
+    EmailsApi::notify_welcome_email(&config, &user).await;
+    Ok(Json(ApiResponse::new(StatusCode::CREATED.into(), user)))
+}
+
+// ❌ Bad — controller handles email error logging
+pub async fn create(...) -> Result<impl IntoResponse, Error> {
+    let user = UserApi::create_by_organization(db, org_id, model).await?;
+    if let Err(e) = EmailsApi::notify_welcome_email(&config, &user).await {
+        warn!("Failed to send welcome email: {e:?}");
+    }
+    Ok(Json(ApiResponse::new(StatusCode::CREATED.into(), user)))
+}
+```
+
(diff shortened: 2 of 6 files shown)


service/src/config.rs
```diff
@@ -143,6 +150,18 @@ pub struct Config {
...
+    #[arg(long, env, default_value = "/coaching-sessions/{session_id}")]
+    session_scheduled_email_url_path: String,
+    /// URL path template for action-assigned email links.
+    /// Use `{session_id}` as a placeholder for the coaching session ID.
+    #[arg(
+        long,
+        env,
+        default_value = "/coaching-sessions/{session_id}?tab=actions"
+    )]
+    action_assigned_email_url_path: String,
```
calebbourg: Having a hard time understanding what this is used for as well.

web/src/controller/organization/user_controller.rs
```diff
@@ -74,7 +74,10 @@ pub(crate) async fn create(
             .await?;
     info!("User created: {user:?}");
 
-    EmailsAPI::notify_welcome_email(&app_state.config, &user).await?;
+    // Best-effort welcome email — log failures, don't block user creation
+    if let Err(e) = EmailsAPI::notify_welcome_email(&app_state.config, &user).await {
```
calebbourg: I'd prefer the logging to happen in `domain` if practical

calebbourg (changes requested):
Left a few comments. The main thing is keeping the controller as thin as possible and moving functions into the `domain` layer

service/src/config.rs
jhodapp: @calebbourg This is used for the scheduled coaching session email template, instead of hardcoding it in the backend code. It's the URL used to be able to click on a link in the email of the newly scheduled coaching session and get navigated to the right place in the frontend.

web/src/controller/organization/user_controller.rs
jhodapp: @calebbourg any particular reason why? There is someone duplicate logging going on between the domain layer (first error logging) and the controller layer (second error logging).

domain/src/gateway/mailersend.rs
```diff
@@ -225,63 +225,58 @@ impl MailerSendClient {
...
     ///
-    /// This method spawns the email sending operation in a background task,
-    /// allowing the caller to continue without waiting for the email to be sent.
-    /// The result of the email sending is logged but not returned to the caller.
-    pub async fn send_email(&self, request: SendEmailRequest) {
-        // Clone values needed for the async task
-        let client = self.client.clone();
-        let base_url = self.base_url.clone();
-        let to_count = request.to.len();
+    /// Errors are returned to the caller so they can be handled by the
```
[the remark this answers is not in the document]
jhodapp: Updated in 17f22aa

calebbourg (approved):

PR #241, opened 2026-03-06
Fix email tests that fail when MailerSend env vars are in .env

jhodapp:
## Description
Email tests that assert missing config values (no API key, no template ID) fail for engineers who have MailerSend env vars in their local `.env` file. The root cause: `Config::new()` calls `dotenv()` internally, which re-reads `.env` and restores vars that the test just removed via `env::remove_var`, causing assertion panics.

The fix separates `.env` loading from `Config` construction:

- **`Config::new()`** — production constructor; reads real CLI args and env vars but no longer calls `dotenv()` itself
- **`Config::default()`** — safe for tests; returns clap defaults via `from_args()` without reading real CLI args or `.env`
- **`Config::from_args([...])`** — new public method for tests that need specific config values
- **`service::load_env_file()`** — explicit `.env` loading, called at application entry points (`main.rs`, `seed_db.rs`)

This also eliminates all uses of the unsafe `env::set_var`/`env::remove_var` from domain tests.

### Changes
* Move `dotenv()` out of `Config::new()` into a new `service::load_env_file()` function
* Call `service::load_env_file()` explicitly in `main.rs` and `seed_db.rs` before `Config::new()`
* Add `Config::from_args()` public method for constructing configs from explicit CLI arg lists
* Redefine `Config::default()` to use `from_args()` — safe, predictable, no side effects
* Replace all `env::set_var` / `env::remove_var` / `EnvGuard` / `Config::default()` in domain email tests with `Config::from_args([...])`
* Remove the `EnvGuard` helper struct from `domain/src/emails.rs`
* Remove all `#[serial]` annotations and the `serial_test` dev-dependency — no longer needed since tests don't share mutable state

### Testing Strategy
* `cargo test -p domain --lib` — all 31 tests pass
* `cargo test -p service --lib` — all 10 tests pass
* `cargo clippy --all-targets -- -D warnings` — clean across entire workspace
* Tests are now deterministic regardless of local `.env` content or shell environment

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -932,7 +932,6 @@ dependencies = [
  "sea-orm",
  "serde",
  "serde_json",
- "serial_test",
  "service",
  "tokio",
 ]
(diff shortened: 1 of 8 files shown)


calebbourg (approved):
Looks great

PR #248, opened 2026-03-11
Implements API integration for Zoom meetings

rnambaale:
## Description
Implements Zoom API integration following the patterns introduced in this PR: https://github.com/refactor-group/refactor-platform-rs/pull/219.

Coaches can connect their Zoom accounts, create Zoom meetings for coaching sessions. Stale OAuth connections cleaned up automatically when tokens are revoked.

### Changes
* Add migrations that add "zoom" to the existing `provider` custom Postgres type
* Migration implements down logic mean't that cleans up data from the `oauth_connections`, and `coaching_sessions` tables should a rollback be required.
* Adds Zoom Oauth 2.0 implementation to `meeting-auth` crate.
* Add Zoom OAuth authorize/callback flow with token storage

### Testing Strategy
* Authenticate a user via `GET /oauth/zoom/authorize` and complete the Zoom OAuth callback
* Verify the connection appears under `GET /oauth/connections` and `GET /oauth/connections/google`
* Call `POST /coaching_sessions/:id/meetings` as the coach with "zoom" as the provider, and confirm the session's meeting_url is populated with a Zoom meeting link
* Call `DELETE /oauth/connections/zoom` and verify the connection row is removed and subsequent `GET /oauth/connections/zoom `returns 404
* Simulate a revoked token scenario (or unit test the invalid_grant path) and confirm HTTP 401 is returned and the stale row is deleted.
* The work here does not break the existing Google integration. The work on a frontend PR that utilises this Zoom integration is underway.

### Deployment Requirements
1. Zoom OAuth credentials and redirect URLs must be configured via environment variables; ensure they are set in all deployment environments before merging.
2. Rename the `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI ` environment variable to `OAUTH_SUCCESS_REDIRECT_URI ` if it was specified in production.

### DIFF
--- a/domain/src/coaching_session.rs
+++ b/domain/src/coaching_session.rs
@@ -3,7 +3,6 @@ use crate::error::{DomainErrorKind, Error, InternalErrorKind};
 use crate::gateway::tiptap::TiptapDocument;
 use crate::Id;
 use chrono::{DurationRound, NaiveDateTime, TimeDelta};
-use entity_api::oauth_connection;
 use entity_api::{
     coaching_relationship, coaching_session, coaching_session_goal, coaching_sessions, mutate,
     organization, query,
@@ -162,15 +161,19 @@ async fn maybe_attach_meeting_url(
     coach_id: Id,
 ) -> Result<(), Error> {
     if let Some(provider) = &coaching_session_model.provider {
-        let has_credentials =
-            crate::oauth_connection::find_by_user_and_provider(db, coach_id, *provider)
-                .await?
-                .is_some();
+        let credentials =
+            crate::oauth_connection::find_by_user_and_provider(db, coach_id, *provider).await?;
 
-        if has_credentials {
-            let meeting_url =
-                create_meeting_url(db, config, coach_id, provider, &coaching_session_model.date)
-                    .await?;
+        if let Some(credentials) = credentials {
+            let meeting_url = create_meeting_url(
+                db,
+                config,
+                coach_id,
+                provider,
+                &coaching_session_model.date,
+                credentials.external_account_id,
+            )
+            .await?;
             coaching_session_model.meeting_url = Some(meeting_url);
         }
     }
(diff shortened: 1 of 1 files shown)


migration/src/m20260311_064303_add_zoom_to_provider_enum.rs
```diff
@@ -0,0 +1,71 @@
...
+            )
+            .await?;
+
+        // 5. Drop the old enum
+        manager
+            .get_connection()
+            .execute_unprepared("DROP TYPE refactor_platform.provider_old")
+            .await?;
+
+        Ok(())
```
calebbourg: Will these steps inadvertently delete any records that use the `google` provider variant?

web/src/controller/oauth_controller.rs
```diff
@@ -129,7 +129,92 @@ pub async fn callback(
...
+        ("code" = String, Query, description = "Authorization code from Zoom"),
+        ("state" = Option<String>, Query, description = "CSRF state token"),
+    ),
+    responses(
+        (status = 302, description = "Redirect to settings page on success"),
+        (status = 400, description = "Invalid callback parameters"),
+        (status = 500, description = "Token exchange failed"),
+    )
+)]
+pub async fn zoom_callback(
```
calebbourg: Let's move this and the `google_callback` function to a new controller called `oauth::callback_controller` and new routes at `oauth/callback/zoom` and `oauth/callback/google`

Let me know if that makes sense

NOTE: If we make this change we will need to update the settings in our google admin to change the callback URL for google oauth. @jhodapp for visibility here as well.

jhodapp (changes requested):
Thanks Raymond, just a few more suggestions that compliment Caleb's review.

domain/src/coaching_session.rs
```diff
@@ -146,6 +154,45 @@ async fn create_meeting_url(
 
             Ok(space.meeting_uri)
         }
+        crate::provider::Provider::Zoom => {
+            let existing_connection = oauth_connection::find_by_user_and_provider(
```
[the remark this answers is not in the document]
rnambaale: @calebbourg 

Just one more thought on this. It looks like this `create_meeting_url` function is only called is when creating a coaching session.

With this recent commit d2a08de657efaacc7707c2fb055c67108181d3dd, the `domain::coaching_session::create` function already makes this same exact query anyway. Would it make sense to refactor the `create_meeting_url` so that it reuses the resulting `oauth_connection` instance from that call? In that case we don't need to query it again here.

PR #249, opened 2026-03-12
Add action→goal FK, carry-forward workflow, and join table queries

jhodapp:
## Description
PR3 of the goals feature: links actions to goals, auto-carries InProgress goals into new sessions, and completes the join table migration started in PR2.

### Changes
* Add nullable `goal_id` FK to `actions` table (ON DELETE SET NULL) with migration and index
* Support `goal_id` query param on `GET /actions` and `GET /users/{id}/actions` for filtering actions by goal
* Implement carry-forward workflow: new coaching sessions auto-link all InProgress goals from the relationship via the join table
* Refactor `batch_load_goals` to query through `coaching_sessions_goals` join table instead of `created_in_session_id`
* Remove obsolete `link_to_originating_session` from domain goal module (CHANGEME markers resolved)
* Update DBML schema definition

### Testing Strategy
* `find_by_user_with_goal_id_filter` — verifies SQL WHERE clause includes `goal_id` filter
* `carry_forward_links_active_goals_to_new_session` — verifies active goals are linked via join table on session creation
* `carry_forward_skips_when_no_active_goals` — verifies no-op when no InProgress goals exist
* All existing tests pass (`cargo test --features mock`)

### Concerns
* Stacked on PR2 (`goals-pr2-relationship-scoping`) — merge PR2 first
* Migration adds a nullable column with no backfill needed (existing actions get `goal_id = NULL`)

### DIFF
--- a/web/src/extractors/compare_api_version.rs
+++ b/web/src/extractors/compare_api_version.rs
@@ -58,7 +58,7 @@ fn is_current_api_version(
 ) -> Result<CompareApiVersion, RejectionType> {
     let version_str = normalize_version(&version);
     let api_version_str = normalize_version(&api_version);
-    warn!(
+    trace!(
         "API version comparison {:?} == {:?}: {}",
         version_str,
         api_version_str,
@@ -67,6 +67,10 @@ fn is_current_api_version(
     if version_str == api_version_str {
         Ok(CompareApiVersion(version))
     } else {
+        warn!(
+            "API version mismatch: client sent {:?}, server expects {:?}",
+            version_str, api_version_str
+        );
         Err((
             StatusCode::BAD_REQUEST,
             format!(


domain/src/coaching_session.rs
```diff
@@ -72,7 +73,23 @@ pub async fn create(
     let tiptap = TiptapDocument::new(config).await?;
     tiptap.create(&document_name).await?;
 
-    Ok(coaching_session::create(db, coaching_session_model).await?)
+    let session = coaching_session::create(db, coaching_session_model).await?;
+
+    // Auto-link all InProgress goals from the relationship to the new session
+    let linked = coaching_session_goal::link_in_progress_goals_to_session(
```
rnambaale: Do you think it would be relevant to wrap this in a database transaction; to rollback the session creation if the auto-linking results in an error?

rnambaale (approved):
This looks good to me. Just one comment / question.

PR #254, opened 2026-03-17
Ai transcription milestone 2

### LINKED ISSUE #261: [Feature]: Milestone 2 - Basic bot transcription
### Suggestion / Feature Request

**Feature Milestone 2**

_First PRs:_
1. Bot joins GMeet (Zoom may be ready to integrate as well) and records the meeting
2. Privacy opt-in/opt-out per user deferred for this milestone
3. Use a Refactor-controlled [Recall.ai](https://recall.ai/) account and auth token
4. Transcription generated and displayed in new tab on coaching session page as a read-only chat-style display
  - This requires using a web hook from the [Assembly.ai](https://assembly.ai/) LeMUR post-processing
  - Use SSE infrastructure to notify the FE

_Second PRs:_
1. Auto-extract Actions from the transcript and associate with the current coaching relationship and session
  - Looking for assignees, due dates
  - Bring your own [Assembly.ai](https://assembly.ai/) account and auth token (needs a way of configuring this from the Settings page)
  - Probably needs a new API tokens table for the backend (double check that Recall or Assembly don't offer OAuth integration)

calebbourg:
## Description                                                                                                                  
   
  Implements the end-to-end meeting recording and transcription pipeline using Recall.ai (Milestone 2 of the AI meeting           
  abstraction layer). When a coach starts a recording, a Recall.ai bot joins the meeting, the recording is automatically
  transcribed with speaker diarization via AssemblyAI (through Recall.ai), and the resulting speaker-labeled transcript segments  
  are stored and served to the coaching session UI. The full pipeline is event-driven via Recall.ai webhooks.

  #### GitHub Issue: Closes [FE Issue #261](https://github.com/refactor-group/refactor-platform-fe/issues/261)                                                                     
   
  ### Changes                                                                                                                     
  * **3 new database migrations** — `meeting_recordings`, `transcriptions`, and `transcript_segments` tables with corresponding
  SeaORM entities and enum types                                                                                                  
  * **Recall.ai gateway** (`domain/src/gateway/recall_ai/`) — HTTP client for bot creation/deletion, async transcript creation,
  transcript metadata retrieval, and pre-signed URL download                                                                      
  * **Domain business logic** — `meeting_recording`, `transcription`, and `transcript_segment` modules encapsulating the recording
   lifecycle (start/stop) and transcription ingestion                                                                             
  * **Entity API layer** — CRUD operations for all three new entities, including status updates and session-scoped lookups
  * **Svix webhook validator** (`meeting-auth/src/webhook/svix.rs`) — HMAC-SHA256 signature verification for Recall.ai webhook    
  payloads                                                                                                                        
  * **Webhook handler** (`web/src/controller/webhook_controller.rs`) — processes `bot.status_change`, `recording.done`, and       
  `transcript.done` events, chaining each stage automatically                                                                     
  * **Three new REST endpoints** (all require auth):
    * `GET|POST|DELETE /coaching_sessions/:id/meeting_recording` — read, start, and stop recording                                
    * `GET /coaching_sessions/:id/transcriptions` — fetch transcription metadata                                                  
    * `GET /coaching_sessions/:id/transcriptions/:tid/transcription_segments` — fetch ordered transcript segments                 
  * **Unauthenticated webhook endpoint** — `POST /webhooks/recall_ai` (Svix-verified)                                             
  * **Config additions** — `RECALL_AI_API_KEY`, `RECALL_AI_REGION`, `RECALL_AI_WEBHOOK_SECRET`, and `WEBHOOK_BASE_URL` env vars   
                                                                                                                                  
  ### Testing Strategy                                                                                                            
  * Set `RECALL_AI_API_KEY`, `RECALL_AI_WEBHOOK_SECRET`, and `WEBHOOK_BASE_URL` in your environment                               
  * Start a coaching session with a real meeting URL (Zoom/Google Meet/Teams)                                                     
  * `POST /coaching_sessions/:id/meeting_recording` — verify a bot joins the meeting and a `meeting_recordings` row is created    
  with status `pending`                                                                                                           
  * Observe status transitions via `GET /coaching_sessions/:id/meeting_recording` as the bot progresses through `joining →        
  in_meeting → recording → processing → completed`                                                                                
  * After recording ends, verify a `transcriptions` row is created and `transcript_segments` are populated once `transcript.done`
  fires                                                                                                                           
  * Verify `DELETE /coaching_sessions/:id/meeting_recording` removes the bot mid-recording
  * Confirm Svix signature rejection with an invalid or missing webhook signature                                                 
                                                                                                                                  
  ### Concerns                                                                                                                    
  * `audio_url` is stored in the DB for internal reference but excluded from API serialization (`#[serde(skip_serializing)]`) —   
  confirm this matches frontend expectations                                                                                      
  * Speaker labels are served as-is from AssemblyAI/Recall.ai with no mapping to user records; the UI will need to handle raw
  labels like `"Speaker A"`                                                                                                       
  * The `WEBHOOK_BASE_URL` must be publicly reachable by Recall.ai during local development (e.g. via ngrok)
  * Recall.ai perfect diarization is only available on Zoom, Google Meet, and Microsoft Teams — Webex and Slack Huddles will get  
  standard (non-separated stream) transcription


docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+Key API calls:
+- `POST /bot` — body includes `meeting_url`, `bot_name`, `webhook_url`, `recording_config`
+- `DELETE /bot/{id}` — stop and remove the recording bot
+
+**`domain/src/gateway/assembly_ai/mod.rs`**
+
+Implements `meeting_ai::traits::transcription::Provider`. Constructed per-request with user's API key.
+
+```rust
+pub struct Provider {
```
jhodapp: I believe that my proposed new Provider enum type makes sense to be added as a provider type in the trait (see longer suggestion comment above).

docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+---
+
+### Step 7: Config Updates
+
+**`service/src/lib.rs`** — add new fields:
+- `recall_ai_api_key: SecretString` — env: `RECALL_AI_API_KEY`
+- `recall_ai_region: String` — env: `RECALL_AI_REGION`, default `"us"`
+- `recall_ai_webhook_secret: SecretString` — env: `RECALL_AI_WEBHOOK_SECRET` (Svix signing secret)
+- `assembly_ai_webhook_secret: SecretString` — env: `ASSEMBLY_AI_WEBHOOK_SECRET` (custom header value)
+- `webhook_base_url: String` — env: `WEBHOOK_BASE_URL` (e.g., `https://app.refactorcoach.com`)
```
jhodapp: Don't we already have a base URL env var we can reuse? I'm worried about having too many static env vars that we could even use existing ones to calculate new ones from.

docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+- `assembly_ai_analysis_model: String` — env: `ASSEMBLY_AI_ANALYSIS_MODEL`, default `"claude-sonnet-4-6"`
```
jhodapp: For a coach that wants to select their own model, would this just act as a default model to use but a coach could override this from their Settings page on the frontend?

jhodapp (changes requested):
Looking great, here's some initial feedback. I did get through the entire thing, but I'm guessing we'll want to discuss some things further that I suggested.

docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+- Recall.ai for recording bots (system-level API key, env var — not per-user)
+- AssemblyAI for transcription + LLM Gateway analysis (per-user API key, stored encrypted in DB)
+- Full webhook automation: each stage triggers the next automatically
+- New tables: `api_credentials`, `meeting_recordings`, `transcriptions`, `transcript_segments`
+- `meeting-manager` crate NOT needed (meeting creation already built in existing gateways)
+- LeMUR is deprecated March 31, 2026 — use AssemblyAI LLM Gateway instead
+- Recall.ai webhooks use Svix HMAC-SHA256 (needs Svix-specific validator added to `meeting-auth`)
+- AssemblyAI webhooks use custom header auth (not HMAC)
+- Providers constructed inline at call site — no AppState provider fields needed
+- `audio_url` is internal-only — not serialized to API clients (`#[serde(skip_serializing)]`)
```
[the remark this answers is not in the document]
jhodapp: Interesting, so they keep the audio and video streams separate for recording?

jhodapp:
@calebbourg Looks like there's a build error that's failing.

domain/src/transcription.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    // Look up the recording to get the bot_id for the Recall.ai API call
+    let recording =
+        recording_api::find_by_bot_id(db, &transcription.meeting_recording_id.to_string()).await?;
+
+    // The bot_id is the Recall.ai "recording ID" used in transcript API paths.
+    let bot_id = recording
+        .as_ref()
+        .map(|r| r.bot_id.clone())
+        .unwrap_or_else(|| transcription.meeting_recording_id.to_string());
```
jhodapp: I think this lookup is pointing at the wrong column. find_by_bot_id runs `WHERE bot_id = $1`, but we're handing it transcription.meeting_recording_id.to_string() — a UUID, not a Recall.ai bot id. So this will always return None, fall through to the unwrap_or_else, and then we'll call `get_async_transcript(&uuid, ...)` on Recall.ai, which will 404.

I think you meant to look up the recording by its primary key. Could we add a `find_by_id` helper in `entity_api::meeting_recording` and drop the fallback? If the FK'd recording isn't there, I'd rather it surface as an error than get silently masked.

web/src/controller/webhook_controller.rs
```diff
@@ -0,0 +1,465 @@
...
+        }
+        Ok(None) => {}
+        Err(e) => {
+            error!(
+                "recording.done: DB error checking transcription for session={}: {:?}",
+                coaching_session_id, e
+            );
+            return StatusCode::INTERNAL_SERVER_ERROR.into_response();
+        }
+    }
```
jhodapp: The idempotency guard in handle_recording_done is a read-then-insert check, and Svix is pretty aggressive about retrying on any 5xx. Two near-simultaneous recording.done deliveries could both see "no transcription" and both call `create_async_transcript`, which means duplicate Recall.ai charges and two transcription rows per session.

Could we add a DB-level guard in the migration in `migration/src/m20260407_000002_add_transcriptions.rs`?

```sql
CREATE UNIQUE INDEX idx_transcriptions_meeting_recording_unique
    ON refactor_platform.transcriptions(meeting_recording_id);
```

Then we can handle the conflict by returning the existing row idempotently. This matches the pattern used in `coaching_sessions_goals_session_goal_unique` and the `actions_users` migrations.

web/src/controller/webhook_controller.rs
```diff
@@ -0,0 +1,465 @@
...
+        }
+    };
+
+    let recording =
+        match MeetingRecordingApi::find_by_bot_id(app_state.db_conn_ref(), &bot_id).await {
+            Ok(Some(r)) => r,
+            Ok(None) => {
+                warn!("recording.done: no recording for bot_id={}", bot_id);
+                return StatusCode::INTERNAL_SERVER_ERROR.into_response();
+            }
```
jhodapp: This match arm returns a 500 Internal Server Error when the recording/transcription can't be found, but Svix treats 5xx as "retry me," and a missing record is a permanent condition and not a transient one. We'll end up with Svix retry storms for up to ~24h on records that will never appear.

Can we return 200 OK here and keep 500 strictly for the Err(e) arm? The log line is enough for alerting.

web/src/controller/webhook_controller.rs
```diff
@@ -0,0 +1,465 @@
...
+    .await
+    {
+        Ok(Some(t)) => t,
+        Ok(None) => {
+            warn!(
+                "transcript.done: no transcription for external_id={}",
+                transcript_id
+            );
+            return StatusCode::INTERNAL_SERVER_ERROR.into_response();
+        }
```
jhodapp: Again here same as above comment, this match arm returns a 500 Internal Server Error when the recording/transcription can't be found, but Svix treats 5xx as "retry me," and a missing record is a permanent condition and not a transient one. We'll end up with Svix retry storms for up to ~24h on records that will never appear.

Can we return 200 OK here and keep 500 strictly for the Err(e) arm? The log line is enough for alerting.

domain/src/meeting_recording.rs
```diff
@@ -0,0 +1,110 @@
...
+        recording.id,
+        MeetingRecordingStatus::Failed,
+        None,
+        None,
+        None,
+        None,
+        Some(chrono::Utc::now().into()),
+        Some("Stopped by user".to_string()),
+    )
+    .await?)
```
jhodapp, line 121: When a user hits stop, we set status to Failed with error_message: "Stopped by user". This conflates "user cancelled" with "bot crashed," and anything reading the status column (UI, analytics, retry logic) can't tell them apart.

I recommend adding a `Cancelled` (or `Stopped`) variant in `entity/src/meeting_recording.rs` and update the migration in place.

web/src/router.rs
```diff
@@ -599,6 +605,44 @@ fn oauth_routes(app_state: AppState) -> Router {
...
+
+fn coaching_session_transcription_segment_routes(app_state: AppState) -> Router {
+    Router::new()
+        .route(
+            "/coaching_sessions/:coaching_session_id/transcriptions/:transcription_id/transcription_segments",
+            get(coaching_session::transcription_segment_controller::index),
+        )
+        .route_layer(from_fn(require_auth))
+        .with_state(app_state)
+}
```
jhodapp, line 698: Heads up: these routes only apply `require_auth`, so any authenticated user can `POST /coaching_sessions/:id/meeting_recording` on sessions they don't participate in, which dispatches a real Recall.ai bot into someone else's meeting.

I know this matches the existing pattern (`POST/DELETE` `session-goals` routes are also unprotected in the same way), so this isn't a regression. But the blast radius is higher here (external side effects, Recall.ai billing, a bot showing up uninvited in another GMeet/Zoom call), so I think it's worth either:

Adding an `is_participant` check in the create handler as defense-in-depth, or
filing a follow-up to add a shared `protect::coaching_sessions::by_id` middleware for all session-nested routes. Up to you on what you think is the best course of action here.

entity_api/src/meeting_recording.rs
```diff
@@ -0,0 +1,236 @@
...
+/// Finds a recording by Recall.ai bot ID — used by webhook handlers
+pub async fn find_by_bot_id(db: &DatabaseConnection, bot_id: &str) -> Result<Option<Model>, Error> {
+    Ok(Entity::find()
+        .filter(Column::BotId.eq(bot_id))
+        .one(db)
+        .await?)
+}
+
+/// Updates recording status and optional artifact fields
+#[allow(clippy::too_many_arguments)]
```
jhodapp: You probably missed this in your own review, consolidate into another struct or two and remove this clippy warning silencing line.

entity_api/src/meeting_recording.rs
```diff
@@ -0,0 +1,236 @@
...
+        coaching_session_id: Unchanged(existing.coaching_session_id),
+        bot_id: Unchanged(existing.bot_id),
+        status: Set(status),
+        video_url: Set(video_url.or(existing.video_url)),
+        audio_url: Set(audio_url.or(existing.audio_url)),
+        duration_seconds: Set(duration_seconds.or(existing.duration_seconds)),
+        started_at: Set(started_at.or(existing.started_at)),
+        ended_at: Set(ended_at.or(existing.ended_at)),
+        error_message: Set(error_message.or(existing.error_message)),
+        created_at: Unchanged(existing.created_at),
```
jhodapp: Small one: `video_url: Set(video_url.or(existing.video_url))` means callers can never clear a field. `None` always means "keep the old value." Other `update_status` functions in `entity_api` (action, goal) are single-field and don't have this ambiguity.

Could we add a doc comment saying "None = preserve existing" so future callers don't get surprised? Or as another option, if we want to allow clears, switch to `Option<Option<T>>`.

entity_api/src/transcription.rs
```diff
@@ -0,0 +1,235 @@
...
+        coaching_session_id: Unchanged(existing.coaching_session_id),
+        meeting_recording_id: Unchanged(existing.meeting_recording_id),
+        external_id: Unchanged(existing.external_id),
+        status: Set(status),
+        language_code: Unchanged(existing.language_code),
+        speaker_count: Unchanged(existing.speaker_count),
+        word_count: Set(word_count.or(existing.word_count)),
+        duration_seconds: Unchanged(existing.duration_seconds),
+        confidence: Set(confidence.or(existing.confidence)),
+        error_message: Set(error_message.or(existing.error_message)),
```
jhodapp, line 117: Small one (same issue as other previous comment): `video_url: Set(video_url.or(existing.video_url))` means callers can never clear a field. None always means "keep the old value." Other `update_status` functions in `entity_api` (action, goal) are single-field and don't have this ambiguity.

Could we add a doc comment saying "None = preserve existing" so future callers don't get surprised? Or as another option, if we want to allow clears, switch to `Option<Option<T>>`.

migration/src/m20260407_000001_add_meeting_recordings.rs
```diff
@@ -0,0 +1,81 @@
...
+                video_url           TEXT,
+                audio_url           TEXT,
+                duration_seconds    INTEGER,
+                started_at          TIMESTAMPTZ,
+                ended_at            TIMESTAMPTZ,
+                error_message       TEXT,
+                created_at          TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+                updated_at          TIMESTAMPTZ NOT NULL DEFAULT NOW()
+            )
+        "#;
```
jhodapp, line 43: Every webhook handler does `find_by_bot_id`, and there's no index or uniqueness constraint on `bot_id`. That means a full table scan per webhook, plus no DB-level guard if a double-create recording bug ever slips in. Could we add:

```sql
CREATE UNIQUE INDEX idx_meeting_recordings_bot_id
    ON refactor_platform.meeting_recordings(bot_id);
```

web/src/controller/webhook_controller.rs
```diff
@@ -0,0 +1,465 @@
+//! Webhook handler for Recall.ai events.
+//!
+//! Response code semantics for Svix retry logic:
+//! - 200 OK: event received and processing started (or idempotent skip)
+//! - 401 Unauthorized: signature invalid; Svix should not retry
+//! - 500 Internal Server Error: DB failure during sync lookup; Svix will retry
```
jhodapp: This comment says "500: DB failure during sync lookup", but as of this implementation we're also returning 500 on `Ok(None)` (see my other comment). I recommend fixing that related code to match the comment, or update the comment once the behavior settles.

web/src/controller/webhook_controller.rs
jhodapp, line 1: Not a blocker and consistent with how other controllers look, but this file is doing a lot: signature validation orchestration, event dispatch, field extraction, idempotency checks, spawn orchestration, and compensating writes. Per coding-standards.md's "thin controllers" guidance, I think a lot of this would fit nicely in a `domain/src/webhook/recall_ai.rs` module or something like it, leaving the controller to just parse → dispatch → status code.

Happy to file as a follow-up if you'd rather not touch it in this PR.

jhodapp (changes requested):
Thanks for this, several pieces of feedback inline plus one here:

`domain/src/transcription.rs`, `domain/src/meeting_recording.rs`: Domain modules on main (emails.rs, goal.rs, coaching_session.rs, etc.) all have tests, but the new `transcription.rs` and `meeting_recording.rs` don't. Worth noting: a test on handle_completion would have caught the find_by_bot_id bug above before it shipped.

Even one happy-path test per module would go a long way, the gateway can be mocked or stubbed.

calebbourg:
@calebbourg please add a flow diagram to this for the recording/processing pipeline. Put in docs

web/src/controller/webhook_controller.rs
jhodapp: @calebbourg quick follow-up — I am not seeing the UNIQUE index on `transcriptions(meeting_recording_id)` yet. The existing `idx_transcriptions_external_id` unique index will not catch this race, because two concurrent `create_async_transcript` calls return different external IDs and both would insert successfully. Could we add the DB-level constraint as originally suggested? Thanks!

web/src/controller/webhook_controller.rs
jhodapp: @calebbourg the controller refactor cleaned this up nicely 👍 — one bit still leaks through though: `domain/src/webhook/transcript_done.rs:24` returns `EntityErrorKind::NotFound` for missing records, which the controller catch-all maps back to 500 at `webhook_controller.rs:56`. Could we have the domain function return `Ok(())` (or a "skip" outcome) for the not-found case so Svix does not retry for ~24h on a permanent miss?

web/src/controller/webhook_controller.rs
jhodapp: @calebbourg same note as on the earlier arm — the refactor removed the explicit 500 here, but `transcript_done` still surfaces `NotFound` to the dispatcher, which the controller maps to 500. A small adjustment in the domain handler should close it.

domain/src/meeting_recording.rs
jhodapp: @calebbourg nice catch on switching user-stop to set status `Processing` rather than `Failed: Stopped by user` — that removes the immediate conflation. The original concern still applies at the terminal state though: once the row lands at `Completed`/`Failed`, we cannot distinguish a user-initiated stop from a natural completion for analytics or the UI. Want to either add the `Cancelled` (or `Stopped`) variant now, or file a follow-up issue?

web/src/router.rs
jhodapp: @calebbourg friendly bump on this — the three new route groups (`meeting_recording`, `transcriptions`, `transcription_segments`) still only have `route_layer(from_fn(require_auth))`. The blast radius is meaningfully bigger than the existing session-goals routes (launching billable Recall.ai bots into arbitrary meetings, cross-session transcript exfiltration), so I would really like at least an `is_participant` check in the create/delete handlers before this merges. Happy to chat about the right shape if useful.

entity_api/src/meeting_recording.rs
jhodapp: @calebbourg small nit still pending — `#[allow(clippy::too_many_arguments)]` is still on `update_status` (9 params). Could we group the optional artifact fields (`video_url`, `audio_url`, `duration_seconds`, `started_at`, `ended_at`, `error_message`) into an `UpdateArtifacts`-style struct so we can drop the allow?

entity_api/src/meeting_recording.rs
jhodapp: @calebbourg tiny follow-up — could we add a doc comment noting that `None` preserves the existing value (i.e. callers cannot clear a field, only overwrite it)? Just to save the next reader from being surprised by the `Set(x.or(existing.x))` semantics.

entity_api/src/transcription.rs
jhodapp: @calebbourg same tiny ask here as on the meeting_recording version — a one-line doc note that `None` preserves the existing value would be great, just so future readers do not get surprised.

migration/src/m20260407_000001_add_meeting_recordings.rs
jhodapp: @calebbourg thanks for adding the index! 🙏 One thing — it landed as `CREATE INDEX IF NOT EXISTS idx_meeting_recordings_bot_id` rather than `CREATE UNIQUE INDEX`. Could we make it unique so a double-create bug cannot slip into prod without Postgres pushing back? That was the half I was most after.

docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+- Meeting creation gateways: `domain/src/gateway/google_meet.rs` + `zoom.rs` functional
+- `meeting_auth::credentials::Storage` trait defined but no DB implementation yet
+
+**What this milestone delivers:**
+End-to-end pipeline: coach starts recording → Recall.ai bot joins meeting → recording completes → AssemblyAI transcribes → LLM Gateway extracts coaching actions → actions persisted automatically. Speaker-diarized transcript segments stored and served for the coaching session conversation UI.
+
+**Key decisions:**
+- Recall.ai for recording bots (system-level API key, env var — not per-user)
+- AssemblyAI for transcription + LLM Gateway analysis (per-user API key, stored encrypted in DB)
+- Full webhook automation: each stage triggers the next automatically
```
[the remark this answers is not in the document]
jhodapp: @calebbourg following up on this — I do not see a watchdog/timeout mechanism for orphaned `Pending`/`Processing` recordings (i.e. the case where the `transcript.done` webhook never arrives). Are we deferring this to a follow-up? If so, could you file an issue so it does not get lost?

docs/implementation-plans/ai-meeting-milestone-2.md
jhodapp: @calebbourg quick check — `WEBHOOK_BASE_URL` is still a separate env var in `service/src/config.rs`. Is there a reason we cannot derive it from an existing base-URL config (so deployers only set one)? Trying to keep the env surface from sprawling. Happy with a short justification if there is a good reason to keep it standalone.

jhodapp (changes requested):
@calebbourg Here's a fresh Rust-quality-centric PR review from Claude. There's enough here that I'm just going to paste it as-is instead of as inline comments.

---
**Blockers**

1. New endpoints lack protect:: middleware — web/src/router.rs:632-662. Compare to coaching_sessions_routes which gates every route. `POST /coaching_sessions/:id/meeting_recording` lets any authenticated user spin up a Recall.ai bot against an arbitrary session with an attacker-supplied `meeting_url`. The GETs leak transcripts and `video_url` cross-session.

2. `transcription_segment_controller::index` ignores `coaching_session_id` — web/src/controller/coaching_session/transcription_segment_controller.rs:35. Even after fixing (1), a user can pass `/coaching_sessions/<A>/transcriptions/<B's-tid>/...` and get B's segments — no check that transcription belongs to the path session.

3. Late recording.failed/bot.fatal overwrites Completed — domain/src/webhook/recording_failed.rs:21-32 & domain/src/webhook/bot_fatal.rs:21-32. No terminal-state guard, unlike bot_status.rs:21-30. Out-of-order delivery silently corrupts session data.

**Major issues**

4. TOCTOU on duplicate recording.done — domain/src/webhook/recording_done.rs:36-45. Concurrent webhooks both pass the `find_by_coaching_session.is_some()` check and both call `create_async_transcript`, creating two transcriptions with different external IDs. Apply the same `try_claim_for_processing` UPDATE pattern used for `transcript.done`.

5. TOCTOU on POST start-recording — web/src/controller/coaching_session/meeting_recording_controller.rs:86-103. Double-submit creates two billable bots. Add a partial `unique index (coaching_session_id) WHERE status NOT IN ('completed','failed')`.

6. `transcript.done` arriving before its row exists returns 500 — domain/src/webhook/transcript_done.rs:16-29. The row is created by a detached `tokio::spawn` in recording_done. If transcript.done wins the race, Svix retries against a 500 until its retry budget expires. `transcript.processing` is a no-op so there's no fallback. Persist in `transcript.processing`, or 200/skip on NotFound.

7. Detached `tokio::spawn` with no panic capture, no `JoinHandle`, holds DB connections across remote downloads — domain/src/webhook/recording_done.rs:77-130, transcript_done.rs:47-80. Bypasses the documented 3-connection-per-request budget; failures vanish silently.

8. `reqwest::Client` rebuilt per request — domain/src/meeting_recording.rs:16-25, domain/src/transcription.rs:22-31. No `.timeout()` either — a hanging Recall.ai endpoint blocks webhook handlers indefinitely. Build one client at startup in AppState.

9. Svix Validator + HMAC key rebuilt per request — web/src/extractors/svix_signature.rs:33. Build once at startup; fail-fast on malformed secret.

10. Body read fully (2 MiB default) before signature verification — [web/src/extractors/svix_signature.rs:51. Unauthenticated DoS surface. Add `DefaultBodyLimit::max(16 * 1024)`.

11. 1unwrap_or(entry)1 in Svix v1, parsing — meeting-auth/src/webhook/svix.rs:122-129. Entries missing the v1, prefix fall back to comparing the whole entry as the signature. Use `let Some(b64_sig) = entry.strip_prefix("v1,") else { continue; };`.

12. `bot.metadata.coaching_session_id` is never cross-checked against `recording.coaching_session_id` — domain/src/webhook/mod.rs:158-163. If anyone with API access to your Recall workspace mints events, they can link arbitrary sessions. Validate equality after `find_by_bot_id`.

12. `audio_url` skip-serialized but `video_url` exposed — entity/src/meeting_recording.rs:53-56. Both are pre-signed Recall URLs with equivalent trust properties. Pick a single policy.

**Minor issues / style**

13. Redundant type prefix — meeting-auth/src/webhook/mod.rs:6-7 re-exports as SvixValidator/HmacValidator. Per .claude/CLAUDE.md, module path is the disambiguator. Callers should use `meeting_auth::webhook::svix::Validator as SvixValidator;`.

14. `status_history: vec![]` discards data — domain/src/gateway/recall_ai/mod.rs:288. `BotDetailResponse.status_changes` already has it.

15. `coalesce_entries` duplicated between domain/src/gateway/recall_ai/mod.rs:188-265 and domain/src/transcription.rs:122-186.

16. 5-minute future timestamp tolerance is too generous — meeting-auth/src/webhook/svix.rs:99 uses `.abs()`. Tighten future skew to ~1 minute.

17. `word_count as i32`, `(seg_start * 1000.0) as i32` truncate silently for long sessions — domain/src/transcription.rs:194, transcription.rs:214-215. Use `i32::try_from(...).unwrap_or(i32::MAX)`.

18. `update_status` can't clear optional fields — entity_api/src/meeting_recording.rs:83-89. `Set(x.or(existing.x))` semantics. Document or switch to a `Patch<T>` enum.

19. Raw SQL with hardcoded refactor_platform. schema — entity_api/src/transcription.rs:55-66. Rest of codebase relies on search_path. Use SeaORM's filtered `update_many`.

20. sentiment_analysis: true is billable and unused — domain/src/gateway/recall_ai/mod.rs:436. sentiment column always `Set(None)` at transcription.rs:217.

**Test coverage gaps**

1. Recall.ai gateway: zero tests. No mockito coverage for error paths, JSON decode failures, timeouts.
2. coalesce_entries: no tests for speaker-change splits, gap-based splits, single-word, empty.
3. State machine: no test that bot_status skips when already Completed/Failed; no test for late recording.failed after recording.done; no test of try_claim_for_processing returning false.
4. Svix validator: no multi-entry signature header test, no webhook-* (post-2025-12-15) header alias test at the extractor level.
5. Recording controller: no test for 409 on duplicate active recording, no test for empty meeting_url, no authorization test (because the authorization doesn't exist).

**Documentation**
1. Missing setup and configuration documentation for this feature geared towards a local developer being able to set up and test this, but also for production deployment set up.

docs/implementation-plans/ai-meeting-milestone-2.md
calebbourg: This was removed

docs/implementation-plans/ai-meeting-milestone-2.md
calebbourg: This isn't needed anymore as it's configured in recall.ai's system 20945349dfb10fd921fb974935b161d478c6d306

web/src/controller/webhook_controller.rs
calebbourg: added! ddd897a5145a1d85671cbc1b8f61df4700a04c41

web/src/controller/webhook_controller.rs
calebbourg: 3d800a1c6a818ec36e918094c3fc538a3821c5b0

web/src/controller/webhook_controller.rs
calebbourg: Made this change 3d800a1c6a818ec36e918094c3fc538a3821c5b0

domain/src/meeting_recording.rs
calebbourg: made added cancelled 250192b1a6d03318b87d96f962c2d4717c65f48f

entity_api/src/meeting_recording.rs
calebbourg: 3d32d537928c357098cb6aa67fe93296a6a49cfe

entity_api/src/transcription.rs
calebbourg: 3d32d537928c357098cb6aa67fe93296a6a49cfe

migration/src/m20260407_000001_add_meeting_recordings.rs
calebbourg: f304bc77ffe7c956e2c0123422d9ff3254fdfc0c

web/src/controller/webhook_controller.rs
calebbourg: b2fb28a6db20d89ad182589f096e2f67a4777e9d

web/src/controller/webhook_controller.rs
calebbourg: b2fb28a6db20d89ad182589f096e2f67a4777e9d

calebbourg:
@jhodapp I believe I have worked through all of your comments. Thank you for the in depth review!

See bcbe53e...52fc182 for addressing your latest batch.

web/src/router.rs
calebbourg: This check is handled by the newer extractor pattern via `CoachingSessionAccess`.

docs/implementation-plans/ai-meeting-milestone-2.md
```diff
@@ -0,0 +1,452 @@
...
+- Providers constructed inline at call site — no AppState provider fields needed
+- `audio_url` is internal-only — not serialized to API clients (`#[serde(skip_serializing)]`)
+- Transcript content lives exclusively in `transcript_segments`; `transcriptions` stores only metadata
+- Speaker labels resolved to real user names via the coaching relationship at segment creation time
+
+---
+
+## Implementation Steps
+
+### Step 1: Database Migrations + SeaORM Entities
```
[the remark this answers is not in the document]
calebbourg: Noting that we agreed to implement as a separate PR

calebbourg:
@jhodapp responses to your latest comments (and things I had missed). I responded to a few inline but listing these here as I'm having a hard time sifting through all of the threads and don't want to miss anything important:
- 6f416fefd70504fa85c60c44cbd78d5399ada6b5  commit for removing the too_many_arguments macro

These will be handled in subsequent PR(s)
- Cost observability proposal + MeetingProvider/PipelineProvider enum split([comment 2962433359](https://github.com/refactor-group/refactor-platform-rs/pull/254#discussion_r2962433359)) — Large proposal from Jim, Caleb never replied. May have been agreed to be out of scope, but it’s not closed.
Pass user UUID through to Recall/Assembly for utterance correlation([comment 2962459778](https://github.com/refactor-group/refactor-platform-rs/pull/254#discussion_r2962459778)) — No response.
Add Provider type to the trait([comment 2962473209](https://github.com/refactor-group/refactor-platform-rs/pull/254#discussion_r2962473209)) — Tied to the proposal above.

I created an issue for the orphaned transaction scenario that you outlined as well. I decided it would be a decent enough lift that it probably should be it's own PR. (Adding cron jobs, retry policies, etc.)


Thank you for the attention to detail on this one!

jhodapp (approved):
Looks amazing! Thank you Caleb.

calebbourg:
Local setup instructions for testing

Meeting Transcription Local Setup

Overview

Meeting transcription requires three sets of credentials: an AES encryption key (for storing OAuth tokens in the database), Google OAuth (so coaches can authorize access to their Google Meet sessions), and Recall.ai (the recording bot + transcription service).

---
1. Encryption Key

All OAuth tokens are encrypted at rest in the database. This key must be set before any OAuth flow works.

ENCRYPTION_KEY=<64-hex-char string>

Generate one:

openssl rand -hex 32

---
2. Google OAuth Setup

Create a Google Cloud Project

1. Go to console.cloud.google.com and create a new project (or use an existing one).
2. Enable these two APIs:
  - Google Meet API (meet.googleapis.com)
  - Google People API (people.googleapis.com)
3. Navigate to APIs & Services → OAuth consent screen:
  - User Type: External
  - Fill in App name, support email, developer contact
  - Add scopes:
      - openid
    - email
    - profile
    - https://www.googleapis.com/auth/meetings.space.created
  - Add your Google account as a Test user (required while the app is in "Testing" status)
4. Navigate to APIs & Services → Credentials → Create Credentials → OAuth 2.0 Client ID:
  - Application type: Web application
  - Authorized redirect URI: http://localhost:4000/api/auth/google/callback
  - Download or copy the Client ID and Client Secret

Environment Variables

GOOGLE_CLIENT_ID=<your-client-id>.apps.googleusercontent.com
GOOGLE_CLIENT_SECRET=<your-client-secret>
GOOGLE_REDIRECT_URI=http://localhost:4000/api/auth/google/callback
OAUTH_SUCCESS_REDIRECT_URI=http://localhost:3000/settings

# These have working defaults — only set if you need to override:
# GOOGLE_OAUTH_AUTH_URL=https://accounts.google.com/o/oauth2/v2/auth
# GOOGLE_OAUTH_TOKEN_URL=https://oauth2.googleapis.com/token
# GOOGLE_USERINFO_URL=https://www.googleapis.com/oauth2/v2/userinfo
# GOOGLE_MEET_API_URL=https://meet.googleapis.com/v2

---
3. Recall.ai Setup

Get an API Key

1. Sign up at recall.ai and create a workspace.
2. In the dashboard, navigate to API Keys and generate a new key.
3. Note your region — use us-east-1 for US workspaces, eu-west-2 for EU.

Configure a Webhook

Recall.ai sends bot lifecycle events (recording done, transcript ready, etc.) to your backend at POST /webhooks/recall_ai. For local development, the backend must be publicly reachable — use ngrok or a similar tunnel.

# Start the backend first, then in another terminal:
ngrok http 4000
# Copy the https URL, e.g. https://abc123.ngrok.io

In the Recall.ai dashboard:

1. Go to Webhooks → Add Endpoint
2. URL: https://<your-ngrok-url>/webhooks/recall_ai
3. Select events: bot.status_change, recording.done, transcript.done
4. After saving, copy the Signing Secret — it starts with whsec_

Environment Variables

RECALL_AI_API_KEY=<your-recall-ai-api-key>
RECALL_AI_REGION=us-east-1          # or eu-west-2
RECALL_AI_WEBHOOK_SECRET=whsec_<base64-encoded-secret>

---
4. Full .env Snippet

Add these blocks to your .env (or .env.local):

# ==============================
#   Encryption (required for OAuth token storage)
# ==============================
ENCRYPTION_KEY=<output of: openssl rand -hex 32>

# ==============================
#   Google OAuth
# ==============================
GOOGLE_CLIENT_ID=<client-id>.apps.googleusercontent.com
GOOGLE_CLIENT_SECRET=<client-secret>
GOOGLE_REDIRECT_URI=http://localhost:4000/api/auth/google/callback
OAUTH_SUCCESS_REDIRECT_URI=http://localhost:3000/settings

# ==============================
#   Recall.ai
# ==============================
RECALL_AI_API_KEY=<recall-api-key>
RECALL_AI_REGION=us-east-1
RECALL_AI_WEBHOOK_SECRET=whsec_<signing-secret>

---
5. Local Development Flow

1. Generate and set ENCRYPTION_KEY.
2. Start the backend: cargo run.
3. Start an ngrok tunnel: ngrok http 4000.
4. Register the ngrok URL as the Recall.ai webhook endpoint (see above).
5. In the frontend, connect Google Meet via the settings page — this triggers the OAuth flow to GOOGLE_REDIRECT_URI.
6. Once connected, starting a coaching session with a Google Meet link will dispatch a Recall.ai bot. Bot events arrive at /webhooks/recall_ai and are verified using RECALL_AI_WEBHOOK_SECRET.

 Note: The ngrok URL changes every time you restart ngrok (on the free plan). You'll need to update the Recall.ai webhook endpoint each session unless you use a stable tunnel URL or a paid ngrok plan.

jhodapp:
> Local setup instructions for testing
> 

@calebbourg Thanks for this, can you please add this as part of a polished setup process under docs/? We could use a new setup.md file that README.md refers to. And include a section for dev setup and production setup. Thanks!

Also in step 1, the generator line for ENCRYPTION_KEY uses 32 bits but the hint wants 64 bits?

calebbourg:
@jhodapp I added the step document in c0e5edecb0abab4a64fed5c0e0ef2444c12f0224 

If we want a more comprehensive setup for the other types of things (tiptap etc.) I'm going to put that out of scope as this has gotten quite large already

PR #273, opened 2026-03-31
Implements magic login links

rnambaale:
## Description
Implements magic login links for new users created within an organization. Instead of requiring the admin to set a password at creation time, new users receive a time-limited magic link in their welcome email. Clicking the link takes them to a setup page where they can set their own password.

### Changes
* **Database migration**: Makes `users.password` nullable and creates a new `magic_link_tokens` table with SHA-256 hashed token storage, expiry timestamps, and a unique constraint on `token_hash`
* **Magic link token lifecycle**: New `domain::magic_link_token` module handles token generation (32 bytes of randomness, URL-safe base64 encoded), validation, and consumption — all within database transactions to prevent concurrent redemption
* **Welcome email integration**: `notify_welcome_email` now generates a magic link token and includes a `magic_link_url` personalization variable in the MailerSend template. The `SessionUrlBuilder` was generalized to support arbitrary URL placeholders
* **New unauthenticated endpoints**:
  - `GET /magic-link/validate?token=...` — validates a token without consuming it, returns user profile data to pre-fill the frontend form
  - `POST /magic-link/complete-setup` — consumes the token, sets the user's password, returns the updated user
* **Password now optional**: `users.password` changed from `String` to `Option<String>`. Users without a password cannot authenticate via email/password until they complete the account setup
* **Config**: Two new environment variables — `MAGIC_LINK_EXPIRY_SECONDS` (default: 72 hours) and `MAGIC_LINK_EMAIL_URL_PATH` (default: `/setup/{token}`)

### Testing Strategy
* All existing tests updated for `Option<String>` password field and new `send_welcome_email` signature
* Welcome email tests verify `magic_link_url` personalization data is included in the MailerSend request body
* Manual testing: create a user via `POST /organizations/:id/users` without a password, verify the welcome email contains a valid setup URL, call `/magic-link/validate` with the token, then `/magic-link/complete-setup` to set a password, and finally log in with the new credentials

### Concerns
* The `magic_link_tokens` table uses application-level enforcement (delete-before-insert) to limit one active token per user, rather than a `UNIQUE(user_id)` database constraint. This allows for future flexibility (e.g., multiple token types) but means stale tokens could accumulate if the application crashes between operations
* The MailerSend template must be updated to include the `{{magic_link_url}}` variable — this is a configuration change outside the codebase
* Existing users with passwords are unaffected; the migration only drops the `NOT NULL` constraint
* The changes here would require a pull request on the frontend repo to remove password fields from the organization member creation form, but also set up the page where the user is routed after clicking the magic link.


calebbourg (approved):

migration/src/m20260330_000000_add_magic_link_tokens.rs
jhodapp, line 1: Don't forget to update docs/db/refactor_platform_rs.dbml and generate a new docs/db/refactor_platform_rs.png by going to dbdiagram.io, dropping the DBML into the digram editor and then exporting a fresh .png.

entity/src/users.rs
```diff
@@ -75,6 +75,6 @@ impl AuthUser for Model {
     }
 
     fn session_auth_hash(&self) -> &[u8] {
-        self.password.as_bytes()
+        self.password.as_deref().unwrap_or("").as_bytes()
```
jhodapp: All users with password = NULL share the same session auth hash (""). This means if one passwordless user's session is invalidated, axum-login can't distinguish their sessions from another passwordless user's. Consider using a per-user fallback like self.id.to_string() instead of empty string:

```rust
fn session_auth_hash(&self) -> &[u8] {
  // Can't return id.to_string() here (lifetime), 
  // but this needs a stable unique fallback
  self.password.as_deref().unwrap_or("").as_bytes()
}
```

This is low-risk today since passwordless users can't log in, but becomes a real issue if magic links ever create sessions directly.

web/src/router.rs
```diff
@@ -476,6 +477,16 @@ pub fn user_session_routes() -> Router {
     Router::new().route("/login", post(user_session_controller::login))
 }
 
+fn magic_link_routes(app_state: AppState) -> Router {
```
jhodapp, line 492: If a magic link expires (after 72h), the user is stuck — there's no endpoint to generate a new token. Consider adding an admin-facing `POST /organizations/{id}/users/{id}/resend-invite` that calls `create_magic_link` + `send_welcome_email` again.

web/src/controller/magic_link_controller.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+#[utoipa::path(
+    post,
+    path = "/magic-link/complete-setup",
+    request_body = CompleteSetupParams,
+    responses(
+        (status = 200, description = "User profile successfully updated", body = User),
+        (status = 400, description = "Password confirmation does not match"),
+        (status = 503, description = "Service temporarily unavailable")
+    )
```
jhodapp: The `#[utoipa::path]` annotation says token comes from `Query`, but the handler actually reads it from the JSON body via `Json<CompleteSetupParams>`. Remove the `params(...)` attribute:

```rust
// Remove this — token is in the request body, not a query param
params(
    ("token" = String, Query, description = "..."),
),
```

domain/src/emails.rs
```diff
@@ -162,10 +190,8 @@ struct SessionUrlBuilder {
 }
 
 impl SessionUrlBuilder {
-    fn build(&self, session_id: &Id) -> String {
-        let path = self
-            .path_template
-            .replace("{session_id}", &session_id.to_string());
+    fn build(&self, placeholder: &str, value: &str) -> String {
```
jhodapp, line 208: The refactoring from build(session_id) to build(placeholder, value) is clean, but the callers now pass raw placeholder strings like "{token}" and "{session_id}". If these ever drift from the config template strings, it'll silently produce broken URLs with unreplaced placeholders. Consider defining these as constants:

```rust
const SESSION_ID_PLACEHOLDER: &str = "{session_id}";
const TOKEN_PLACEHOLDER: &str = "{token}";
```

web/src/params/user/mod.rs
```diff
@@ -87,3 +87,29 @@ impl IntoUpdateMap for UpdatePasswordParams {
         update_map
     }
 }
+
```
jhodapp, line 90: `struct CompleteSetupParams` is missing `#[derive(ToSchema)]`, so the OpenAPI `request_body = CompleteSetupParams` reference in the controller won't generate docs correctly.

domain/src/magic_link_token.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let raw_token = params.remove("token")?;
+
+    if password != confirm_password {
+        warn!("Password confirmation does not match during magic link setup");
+        return Err(Error {
+            source: None,
+            error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(
+                "Password confirmation does not match".to_string(),
+            )),
```
jhodapp: Password mismatch is a client validation error, not an internal error. This likely maps to a 500 response when it should be a 400. Verify how `InternalErrorKind::Other` maps in the web error handler — it may need a dedicated `DomainErrorKind::Validation` variant or similar.

The error tree domain/src/error.rs currently has no domain-level "client validation error" concept. Everything is either Entity (DB-originated), Config, or Other (catch-all 500). This works fine when validation happens at the entity_api layer (where `InvalidQueryTerm` → `EntityErrorKind::Invalid` → 422), but the magic link complete_setup does validation in the domain layer before touching the DB — so there's no clean path to a 4xx response.

The quickest fix is `EntityErrorKind::Invalid`. The proper fix would be adding something like `InternalErrorKind::Validation(String)` and mapping it to 400/422 in the web error handler — which would benefit any future domain-level validation too.

jhodapp (changes requested):
Thanks for this Raymond. I did a security review with Claude and am pasting some of the ones that I think are probable issues.

We should add some test coverage to this behavior since it's a critical security related process. Here's some suggestions that I came up with with Claude's analysis:

**Token security (domain layer):**

Expired token is rejected (validates the Utc::now() > expires_at check)
Invalid/garbage token returns not-found, not a crash
Token is single-use — complete_setup deletes it, second call fails
Creating a new token invalidates the previous one
hash_token produces consistent SHA-256 output (round-trip test)

**Password validation (domain layer):**

Mismatched password / confirm_password returns an error
Successful setup actually hashes and stores the password

**Auth boundary (entity_api layer):**

authenticate_user rejects login when password is None
verify_password with None hash returns RecordUnauthenticated

**HTTP layer (integration):**

GET /magic-link/validate with valid token returns 200 + user
GET /magic-link/validate with expired token returns 401
GET /magic-link/validate with bogus token returns 404
POST /magic-link/complete-setup with valid token returns 200
POST /magic-link/complete-setup with mismatched passwords returns error (and verify it's not 500)
POST /magic-link/complete-setup after token already consumed returns error
Both endpoints are accessible without an authenticated session (no auth middleware)

migration/src/m20260330_000000_add_magic_link_tokens.rs
rnambaale: Resolved in 056242c841361d93ab882beecddb3e844043a13c.

entity/src/users.rs
rnambaale: Fixed in c48657bf1381d35b465c26b164d436ac6a5ba689.

web/src/router.rs
rnambaale: Added in f3bdc69da7429d6d60b0af9326b3c134b9b0f1bd.

domain/src/emails.rs
rnambaale: Fixed in af87e8dab7b72641e0445dbb960ccfcad1ec4e7b.

web/src/params/user/mod.rs
rnambaale: Fixed in 813037acc19a9011b0fccb0cc1d073fb1c0dee35.

domain/src/magic_link_token.rs
rnambaale: I chose the option of creating a domain-level validation variant. Implemented bcb606d4d7496f6c2efc2689a91ffacfa4feeba0.

Its not used in the when checking for password confirmation in `magic_link_token`, as well as  `user` domains. a7f0569f193b2e31f3672a10daf812c977f77244

jhodapp: @calebbourg would love your thought on this new domain level error kind as an architectural decision. If you're good with it, I am too.

calebbourg: @jhodapp @rnambaale I think that approach works! Thank you for looping me in.

web/src/controller/organization/user_controller.rs
```diff
@@ -72,11 +72,56 @@ pub(crate) async fn create(
...
+        return Err(domain::error::Error {
+            source: None,
+            error_kind: domain::error::DomainErrorKind::Validation(
+                "User has already completed setup".into(),
+            ),
+        }
+        .into());
+    }
+
+    EmailsAPI::notify_welcome_email(app_state.db_conn_ref(), &app_state.config, &user).await;
```
jhodapp: I know in other controllers where we send emails we do ignore errors. But in this case, failing to resend the invite email is a critical error to deal with or at least ensure is being logged as an ERROR.

jhodapp (changes requested):
Great enhancements, thanks for these. Still a couple of things to address:
1. Silent failure of resending an invite email.
2. The follow tests I suggested to add are still missing:
  - Token is single-use (complete_setup deletes it, second call fails)
  - Password mismatch returns validation error (not 500)
  - authenticate_user rejects login when password is None
  - All HTTP-layer integration tests (valid/expired/bogus token via HTTP, complete-setup with mismatched passwords, etc.)

web/src/controller/organization/user_controller.rs
rnambaale: `notify_welcome_email` by design implements a best-effort contract. There was an option of changing this so that is actually returns a result; but that would mean changing the standards.

For now, I have chosen to expose a new `create_and_send_welcome_email` function that reflects the required behaviour for re-sending an invite. Implemented in: 2512a8f82caeadbcfad4ded12e7b2988eb3d426c

I'm happy to iterate on this further.

rnambaale:
> Great enhancements, thanks for these. Still a couple of things to address:
> 
> 1. Silent failure of resending an invite email.
> 2. The follow tests I suggested to add are still missing:
> 
> * Token is single-use (complete_setup deletes it, second call fails)
> * Password mismatch returns validation error (not 500)
> * authenticate_user rejects login when password is None
> * All HTTP-layer integration tests (valid/expired/bogus token via HTTP, complete-setup with mismatched passwords, etc.)

@jhodapp I have added the tests you suggested.

2d2e7b9e76f7a60ea8ce82fee2d3ab004c974733...22f7f0083aeafae959e5c48e7ab16a5c32c58ddf

I couldn't find an existing pattern for HTTP-layer integration tests in controllers. But I have introduced one for the `magic_link_controller` here fdae02508f1120daebe2e20f9e457f806a924b46.

jhodapp (approved):
This is excellent now, thank you @rnambaale. Approved.

PR #281, opened 2026-04-16
Add `invite_status` to user model

rnambaale:
## Description

Add a computed `invite_status` field to the organization users list response so the frontend can display whether each user's account setup is pending, expired, or active.

The status is derived at query time from existing data — no new database migration needed. This was chosen over a stored `invite_status` column approach because:
- It avoids a new migration with an enum type and `ALTER TABLE`
- It eliminates the need to keep a stored column in sync across multiple code paths (user creation, token expiry, setup completion)
- The source of truth already exists: `password` presence and `magic_link_tokens.expires_at`

**Status logic:**
- `active` — user has a password (completed setup)
- `pending` — no password, token exists and has not expired
- `expired` — no password, token is missing or expired

### Changes

* Add `InviteStatus` enum (`Pending`, `Expired`, `Active`) in `entity/src/user_invite_status.rs`
* Add computed `invite_status: Option<InviteStatus>` field to `users::Model` (ignored by SeaORM, serialized only when present)
* Add `find_by_user_ids` batch query in `entity_api/src/magic_link_token.rs` to avoid N+1 queries
* Add `compute_invite_status` pure function in `domain/src/magic_link_token.rs`
* Promote `find_by_organization` from a re-export to a custom domain function that batch-fetches tokens and populates `invite_status`

### Testing Strategy

* `cargo clippy -- -D warnings` passes
* `cargo test` passes
* Verify `GET /organizations/:id/users` response includes `"invite_status"` field with correct values:
  - User with password set returns `"active"`
  - Newly invited user (token not expired) returns `"pending"`
  - User whose token expired returns `"expired"`

### Concerns

* `invite_status` is `null` in responses from endpoints that don't compute it (e.g. login, profile fetch, complete-setup). The frontend should only rely on this field in the organization users list context.
* Each user can have at most one magic link token (enforced by `create_magic_link` which deletes existing tokens before creating a new one), so the batch query returns a simple 1:1 mapping.


entity/src/user_invite_status.rs
jhodapp, line 1: @rnambaale Is this only meant to track invite status of a user, or is it really a shadow way of saying whether a user is active in the system or not? We might want to think through this in relation to [this issue](https://github.com/refactor-group/refactor-platform-fe/issues/214).

If you think they compliment each other as separate things, I'm open to that design but it's worth thinking through here.

jhodapp (commented):
@rnambaale I left you one critical design consideration for now. I like the way you're thinking about this, and once we get clear on the right answer to my question, we'll know how to proceed here.

entity/src/user_invite_status.rs
rnambaale: @jhodapp This is only meant to track invite status of a user. 

I think what we have in the [linked issue](https://github.com/refactor-group/refactor-platform-fe/issues/214) would need a separate kind of statuses. There is an option to merge the two status fields but then that could potentially violate the single-responsibility principle.

jhodapp: @rnambaale I'm down with that, and if we do ever decide to bring them together in the future, it'd be easy enough to do so since this is a computed field on the User and not part of a DB schema migration.

jhodapp (approved):

domain/src/user.rs
```diff
@@ -28,6 +30,28 @@ where
...
+
+    for user in &mut users {
+        user.invite_status = Some(magic_link_token::compute_invite_status(
+            &user.password,
+            token_map.get(&user.id).copied(),
+        ));
+    }
+
+    Ok(users)
+}
```
jhodapp: Auth hot path now pays for invite-status enrichment it doesn't use. `OrganizationMemberAccess` — which runs on every request to `/organizations/{id}/*` — calls this function purely to answer "is the authenticated user a member of this org?" It doesn't read `invite_status`. But after this change, every one of those requests also does:

```rust
let tokens = entity_api::magic_link_token::find_by_user_ids(db, &user_ids).await?;
```

Two problems:

1. **Extra query on every authorized request.** For an org with N members, this is an additional SELECT against magic_link_tokens on every sidebar load, every coaching-session fetch, etc. The auth extractor should stay cheap.
2. **Coupling failure modes.** Because this is wired into auth, a broken/missing magic_link_tokens table takes down authorization for the entire org scope — not just the Members listing that actually needs invite status.

My suggestion: keep `entity_api::user::find_by_organization` as-is (cheap, role-only), and move the token lookup into either:

- a new domain::user::find_by_organization_with_invite_status called only from web/src/controller/organization/user_controller.rs:38, or

- an explicit `enrich_with_invite_status(db, &mut users)` helper the Members controller invokes after the base call.

Then `OrganizationMemberAccess` can call the plain version and stay independent of the magic-link feature.

domain/src/user.rs
```diff
@@ -28,6 +30,28 @@ where
...
 }
 
+pub async fn find_by_organization(
+    db: &DatabaseConnection,
+    organization_id: Id,
+) -> Result<Vec<users::Model>, Error> {
+    let mut users = user::find_by_organization(db, organization_id).await?;
+
+    let user_ids: Vec<Id> = users.iter().map(|u| u.id).collect();
+    let tokens = entity_api::magic_link_token::find_by_user_ids(db, &user_ids).await?;
```
jhodapp, line 44: When users is empty (org has no members which is unusual but possible mid-migration), this still runs `find_by_user_ids(db, &[])`, which most SeaORM backends translate into `WHERE user_id IN ()` — harmless, but worth an early if `users.is_empty() { return Ok(users); }` to avoid the round-trip.

jhodapp (changes requested):
Found a couple of issues when I was testing end-to-end.

domain/src/user.rs
rnambaale: Fixed in: 438847180fdf688db9e1e08c119a7df0e124417e.

domain/src/user.rs
rnambaale: Fixed in 1dc563fb3017b5511f09f3fb7343e98779db0d41.

jhodapp (approved):
Looks great now! Approved.

PR #286, opened 2026-04-21
Relay correct status code for DB errors in OrganizationMemberAccess extractor

### LINKED ISSUE #284: [Bug]: OrganizationMemberAccess extractor misattributes all DB errors as "Organization not found" and returns 400 instead of 500
## Summary

The `OrganizationMemberAccess` extractor at [web/src/extractors/organization_member_access.rs#L67-L77](https://github.com/refactor-group/refactor-platform-rs/blob/main/web/src/extractors/organization_member_access.rs#L67-L77)
treats every `Err(_)` from `UserApi::find_by_organization` as if the organization did not exist:

```rust
let user_organization_role_exists =
    match UserApi::find_by_organization(state.db_conn_ref(), organization_id).await {
        Ok(users) => users.iter().any(|user| user.id == authenticated_user.id),
        Err(_) => {
            error!("Organization not found with ID {organization_id:?}");
            return Err((
                StatusCode::BAD_REQUEST,
                "Invalid organization ID".to_string(),
            ));
        }
    };
```

Two problems:

1. **Misleading log.** `Err(_)` discards the underlying error — connection
   drops, missing tables, timeouts, and serialization failures all get logged
   as `"Organization not found with ID {id}"`, even when the organization
   clearly exists. On-call engineers have no way to see the real Postgres/DB
   error from the logs.

2. **Wrong HTTP status.** A DB-layer failure is not a client error. Returning
   `400 "Invalid organization ID"` sends callers looking at their request
   payload when the real problem is server-side. The `400` branch only makes
   sense for the parse failure a few lines above; DB errors should surface
   as `500`.

## Impact

- Debugging is measurably harder: every DB-layer failure on any
  `/organizations/{id}/*` route produces the same misleading line with no
  underlying error context.
- Clients (frontend, integrators) receive a `400` for server-side incidents,
  which breaks retry logic and alerting expectations.

Encountered in practice on PR
[#281](https://github.com/refactor-group/refactor-platform-rs/pull/281):
a new code path in that PR started querying `magic_link_tokens` during
`find_by_organization`. When the corresponding migration hadn't been applied
locally, every org-scoped request failed, and the logs blamed a valid
organization instead of surfacing the real
`relation "refactor_platform.magic_link_tokens" does not exist`.

## Proposed fix

Log the underlying error and return the correct status:

```rust
Err(err) => {
    error!(
        "find_by_organization({organization_id:?}) failed while verifying membership: {err:?}"
    );
    return Err((
        StatusCode::INTERNAL_SERVER_ERROR,
        "Failed to verify organization membership".to_string(),
    ));
}
```

Optional follow-up: if we want to preserve a `404` for the
"organization genuinely doesn't exist" case, the extractor should match on
the `Error`/`EntityApiErrorKind` variant rather than a blanket `Err(_)`.

## Scope

Pre-existing since the auth-extractor refactor landed on `main`
(commit `87c9ec9`). Not introduced by any open PR, but worth fixing
independently.

rnambaale:
## Description
The `OrganizationMemberAccess` extractor was mapping any failure from `UserApi::find_by_organization` to `400 Bad Request "Invalid organization ID"`.

That response is misleading: the organization ID is already parsed and validated as a UUID earlier in the extractor, and `find_by_organization` returns `Ok(vec![])` (not an error) when no users match — so the `Err(_)` arm can only be reached via a genuine database failure. Infrastructure errors should surface as `500 Internal Server Error`, not as a client-side validation failure, and the log line should reflect the actual cause rather than a fabricated "Organization not found" message.

### Changes
* In `web/src/extractors/organization_member_access.rs`, replace the `400 Bad Request "Invalid organization ID"` response in the DB-error arm with `500 Internal Server Error "Failed to verify organization membership"`.
* Log the underlying `sea_orm` error (`find_by_organization({organization_id:?}) failed while verifying membership: {err:?}`) instead of the misleading "Organization not found with ID ..." message.

### Testing Strategy
* `cargo check -p web` and `cargo clippy` both pass.
* Existing extractor tests (`test_extractor_returns_200_for_users_with_organizational_roles`, `test_extractor_returns_200_for_super_admin_users_without_organizational_roles`, `test_extractor_returns_401_organization_for_regular_users_without_organization_roles`) continue to exercise the success and unauthorized branches; no behavior change for those cases.
* To manually verify the new 500 path, point the backend at an unreachable database (e.g. stop Postgres or set a bad `DATABASE_URL`) and hit a route guarded by this extractor as a non-SuperAdmin user — the response should now be `500 Internal Server Error` with the real `sea_orm` error captured in the logs, rather than a confusing `400`.

### Concerns
* Behavior for a *valid UUID that does not exist in the database* is unchanged: non-admins still receive `401 Unauthorized` (since the member lookup returns an empty set), and SuperAdmins still pass the extractor and fail later in the handler. If we want missing-org requests to return `404` explicitly, that would be a separate change that adds an `organizations::find_by_id` check before the membership lookup.


jhodapp:
@rnambaale What do you think about the stated concern in the PR description, specifically the last line? "If we want missing-org requests to return 404 explicitly, that would be a separate change that adds an organizations::find_by_id check before the membership lookup."

Perhaps this is also worth adding here, as it does make sense to return a 404 if a requested Org's ID isn't found.

rnambaale:
> Perhaps this is also worth adding here, as it does make sense to return a 404 if a requested Org's ID isn't found.

@jhodapp I agree, I think it is a more idiomatic approach. I had not gone that route because of that extra `organizations::find_by_id` lookup; but I think it is quick enough, and the benefits outweigh the cost. 

I have iterated on this more here: 9eba8a571c86a5c179711e10823a1cf6a7ad1118.

jhodapp:
> > Perhaps this is also worth adding here, as it does make sense to return a 404 if a requested Org's ID isn't found.
> 
> 
> 
> @jhodapp I agree, I think it is a more idiomatic approach. I had not gone that route because of that extra `organizations::find_by_id` lookup; but I think it is quick enough, and the benefits outweigh the cost. 
> 
> 
> 
> I have iterated on this more here: 9eba8a571c86a5c179711e10823a1cf6a7ad1118.

This looks great, I'm satisfied. I'll approve.

jhodapp (approved):

PR #291, opened 2026-04-24
ci: replace PR preview dispatch dropdowns with runtime ref resolver

### LINKED ISSUE #290: CI: replace choice-based PR preview dispatch with runtime ref resolver (stop all bot commits to main)
## Problem

`git log main --oneline -30` on this repo shows **17 of the last 30 commits** are `chore: refresh preview commit choices [skip ci]` commits authored by `github-actions[bot]`. Every PR lifecycle event (`opened` / `reopened` / `closed`) fires `.github/workflows/refresh-preview-commits.yml`, which rewrites the `options:` array in `dispatch-pr-preview.yml` via `yq` and pushes the result to `main` using `GHCR_PAT`.

The bot commits:
- Bury real work in `git log`, `git blame`, and changelogs.
- Create the false impression that contributors' PR pushes mutate `main`.
- Show up in release notes.

## Root cause

`type: choice` inputs in a `workflow_dispatch` trigger are rendered server-side from the YAML on the default branch. GitHub provides no hook to populate `choices` dynamically at dispatch-render time. The only way to keep the list fresh is to edit the YAML → commit → push. Therefore, `type: choice` is fundamentally incompatible with "zero bot commits to main."

The `pull_request: synchronize` event is excluded from the trigger, so a raw PR-branch push does not fire the workflow directly — but any subsequent `opened` / `reopened` / `closed` event on any PR will pick up the new `headRefOid` and produce another bot commit. That's the root of the "push → main commit" correlation.

## Fix

1. Replace the `type: choice` inputs on `dispatch-pr-preview.yml` with simple string inputs:
    - `backend_pr_number` — required string (e.g. `289` or `PR#289`).
    - `backend_sha_override` — optional, override the PR branch HEAD.
    - `frontend_ref` — optional, default `main`; accepts `main`, a branch name, a SHA, or `PR#<num>`.
2. Add a resolver step at the top of the validate job that resolves each input to a concrete SHA + branch using the GitHub API (`gh pr view`, `gh api repos/.../commits/<ref>`).
3. **Delete `.github/workflows/refresh-preview-commits.yml`** — no longer needed.
4. Add a new read-only helper workflow `.github/workflows/list-preview-refs.yml` that writes a Markdown table of current `main` commits + open PR heads from both repos to `\$GITHUB_STEP_SUMMARY`. It uses `permissions: contents: read`, never pushes, never commits.
5. Update `.claude/CLAUDE.md` "Commit Choice Refresh" section to describe the new pattern.

## Non-regressions

- Deploying `main/main` → just leave `frontend_ref=main` (default) and pass the backend PR number.
- Deploying an open PR → type `PR#<num>` in the primary input.
- Deploying an exact SHA → pass it in `backend_sha_override`.
- Cross-repo pair selection preserved via the `frontend_ref` input.
- The reusable `ci-deploy-pr-preview.yml` is unchanged — still called with `backend_sha` / `frontend_sha` from the resolver outputs.

## Gotchas / trade-offs

- UX shift: dropdown → text input. Users run `List Preview Refs` (one click) or look at the PR URL if they don't remember the number.
- `GHCR_PAT` stays on the dispatch workflow (unchanged); the new `list-preview-refs.yml` only reads, so it could later downgrade to `GITHUB_TOKEN` (out of scope for this change).
- Cross-repo issue: `refactor-group/refactor-platform-fe` needs a mirrored fix — tracked in a sibling issue.

## Expected outcome

- Zero `github-actions[bot]` commits on `main` going forward (from this workflow).
- `gh run list --workflow refresh-preview-commits.yml --event pull_request` returns empty (the workflow file is deleted).
- PR preview deploys continue to work end-to-end with unchanged behavior for users.

lmcdonough:
Closes #290.

## Context

`git log main --oneline -30` shows **17 of the last 30 commits** are `chore: refresh preview commit choices [skip ci]` commits by `github-actions[bot]`. Every PR lifecycle event (`opened` / `reopened` / `closed`) fires `refresh-preview-commits.yml`, which rewrites `dispatch-pr-preview.yml`'s `options:` array via `yq` and pushes to `main` with `GHCR_PAT`. That pollutes `main` history, invalidates `git blame`, and falsely correlates contributors' PR-branch pushes with `main` mutations.

## Root cause

`type: choice` inputs on a `workflow_dispatch` trigger are rendered server-side from the YAML on the default branch. GitHub provides no hook to populate `choices` dynamically at dispatch-render time. The only way to keep the list fresh is to edit the YAML → commit → push. Therefore `type: choice` is fundamentally incompatible with "zero bot commits to main."

## Fix

- **Replace dropdowns with three text inputs** on `dispatch-pr-preview.yml`:
  - `backend_pr_number` — required (accepts `289` or `PR#289`).
  - `backend_sha_override` — optional, override the PR branch HEAD.
  - `frontend_ref` — optional, default `main`; accepts `main`, a branch name, a 7+ char SHA, or `PR#<num>`.
- **Resolve inputs at dispatch time** via `gh pr view` / `gh api repos/.../commits/<ref>`. Nothing is written to `main`.
- **Add `list-preview-refs.yml`** — a read-only helper (`permissions: contents: read`) that prints current `main` commits + open PR heads from both repos to `\$GITHUB_STEP_SUMMARY`. No commits, no pushes.
- **Delete `refresh-preview-commits.yml`** — it exists only to keep the dropdowns fresh.
- **Update `.claude/CLAUDE.md`** to document the new flow.

Diffstat: `4 files changed, 176 insertions(+), 251 deletions(-)` — net code reduction.

## Non-regressions

| Capability | How it still works |
|---|---|
| Deploy backend PR against `main` | `backend_pr_number=<num>`, `frontend_ref=main` (default) |
| Deploy backend PR against a frontend PR | add `frontend_ref=PR#<num>` |
| Deploy specific SHAs | `backend_sha_override=<sha>` / `frontend_ref=<sha>` |
| Draft PRs | `PR_STATE != OPEN` gate still allows drafts |
| Cross-repo pair selection | `frontend_ref` still queries the frontend repo |
| Cleanup on PR close | Unchanged |
| Reusable `ci-deploy-pr-preview.yml` | Unchanged — still called with `backend_sha` / `frontend_sha` / `pr_number` / `branch_name` |

## UX change (soft)

Instead of a dropdown, users type a short string. For the common "deploy my backend PR against current main," input is `289` (or `PR#289`) — done. For "I need to pair with frontend PR 373," add `frontend_ref=PR#373`. If a user doesn't remember a ref, they run **List Preview Refs** (one click) which prints a Markdown table of every `main` commit and open PR in both repos.

## Test plan

Already validated locally:
- `actionlint .github/workflows/dispatch-pr-preview.yml .github/workflows/list-preview-refs.yml` → clean.
- `act pull_request -l` → **no match** for `refresh-preview-commits` or `Refresh Preview Commits` (proof that PR events can no longer produce a main commit from these workflows).
- `act workflow_dispatch -W dispatch-pr-preview.yml --input backend_pr_number=289 --input frontend_ref=main -n` → dry-run succeeds end-to-end.
- `act workflow_dispatch -W list-preview-refs.yml -n` → dry-run succeeds.
- Input parsing smoke test: 13 assertions covering `\"289\" | \"PR#289\" | \"PR#1\" | \"main\" | \"feature/foo\" | hex SHAs of varying length | 7-hex branch name` all classify correctly.

Post-merge GitHub-side smoke:
1. `gh workflow run list-preview-refs.yml --ref main` → confirm Markdown table appears in the run summary.
2. `gh workflow run dispatch-pr-preview.yml -f backend_pr_number=<live-PR> -f frontend_ref=main` → confirm deploy end-to-end.
3. `gh run list --workflow refresh-preview-commits.yml` → empty (file is gone).
4. 48h observation: `git log --author='github-actions\[bot\]' main --since=<merge-date>` → empty.

## Gotchas

- Mirror issue / PR on the frontend repo: refactor-group/refactor-platform-fe#378. Both changes land independently; each removes its own repo's refresh workflow and dropdowns.
- `GHCR_PAT` is no longer referenced by either new/edited workflow — they use `GITHUB_TOKEN` for API reads. A future cleanup could narrow secret scopes further; out of scope for this PR.


jhodapp (commented):
Looking @lmcdonough. I've included the review I did with Claude.

LGTM overall — right architectural call, net deletion, secret scope tightened. A few non-blocking notes:

**1. Dead `elif` branch in frontend resolver.** `.github/workflows/dispatch-pr-preview.yml` lines ~127–145 — the SHA path and branch path both call `gh api repos/$FE_REPO/commits/$INPUT_FRONTEND_REF` with no behavioral difference (the endpoint accepts branches, tags, and SHAs interchangeably). Only the notice message differs. Suggest collapsing to a single non-`PR#` path.

**2. `backend_sha_override` isn't verified against the PR branch.** Any commit in the repo is accepted, even one that's only on `main` or an unrelated branch — but the deploy still gets tagged with the PR's `branch_name` / `pr_number`. Pre-existing behavior, fine as a power-user escape hatch, but worth one sentence in the workflow header noting "override is unverified against the PR branch."

**3. No checked-in test for the input regex.** The 13-assertion smoke test in the description is ad-hoc local. The classification logic (`^PR#([0-9]+)$`, `^[0-9]+$`, `^[0-9a-fA-F]{7,40}$`) is exactly what silently breaks on future edits. A small bats/bash script under `.github/workflows/tests/` triggered on workflow changes would close the gap.

**4. Helper docs reference frontend-side inputs that don't exist yet here.** `.github/workflows/list-preview-refs.yml` lines ~62–66 tell users to pass `frontend_pr_number` / `backend_ref` to the frontend repo's dispatch — which lives in refactor-group/refactor-platform-fe#378. If this lands first, the instructions are aspirational. Consider hedging the wording or linking the FE PR.

Looks like there's also a minor set of merge conflicts to resolve on the branch compared to `main`.

lmcdonough:
Thanks for the review @jhodapp. Addressed everything in 7031a75.

**1. Dead elif branch.** I collapsed the SHA arm and the branch arm into one fall-through. The gh api commits/<ref> endpoint accepts SHA, branch, or tag interchangeably, so the split arms only differed in their ::notice:: wording. The single arm now logs the resolved short SHA without claiming to know whether the input was a SHA or a branch.

**2. backend_sha_override unverified against PR branch.** Added a header note in dispatch-pr-preview.yml stating the override accepts any commit reachable in the repo and is intentional as a power-user escape hatch. Verifying ancestry would require a checkout per dispatch, which is disproportionate for a manual workflow.

**3. No checked-in test for the input regex.** Added a 50-case bats suite at .github/workflows/tests/dispatch-pr-preview.bats covering PR# precedence, numeric PRs, SHA overrides (length, hex-only, case), branch names with slashes/dots/dashes, hex-named branches, and the contract that the ref classifier never returns INVALID. A new test-dispatch-regex.yml workflow runs bats on every PR that touches the dispatch workflow or the test file, so a future regression now fails review rather than dispatch.

**4. list-preview-refs.yml frontend instructions.** Hedged the line that points at the frontend dispatch with \"Available once refactor-group/refactor-platform-fe#379 lands.\" If this PR lands first the wording is honest about the dependency.

**5. Merge conflicts.** Resolved by taking HEAD wholesale. The main-side updates were dropdown options on a feature this PR removes (refresh-preview-commits.yml is deleted on this branch).

**Bonus from CI run 25157726565 step 11.** I tracked down the 404 on the \"Comment on PR with Preview URLs\" step. The reusable ci-deploy-pr-preview.yml was using context.repo when it should derive the target repo from inputs.repo_type. When the FE dispatch invoked the BE reusable workflow with repo_type=frontend the comment tried to post to refactor-platform-rs#373 instead of refactor-platform-fe#373. Fixed by deriving targetOwner/targetRepo from inputs.repo_type and switching the github-script step to GHCR_PAT (the default GITHUB_TOKEN is single-repo scoped and cannot comment cross-repo).

.github/workflows/test-dispatch-regex.yml
jhodapp, line 1: @lmcdonough Just a heads up: adding `test-dispatch-regex.yml` means another entry in the repo's Actions sidebar (it'll show up as "Test Dispatch Regex"). GitHub's sidebar is file-based, there's no hidden: flag, so every workflow file gets its own row whether or not it's meant for humans to click.

A few options if we want to keep the list tight:

1. Consolidate — fold the bats job into `dispatch-pr-preview-frontend.yml` as an additional job gated by paths:. One file, one sidebar entry, two jobs. Same trick could apply to `cleanup-pr-preview-frontend.yml`.
2. Drop `workflow_dispatch`: on workflows that aren't meant to be human-triggered, doesn't remove the entry but kills the "Run workflow" button so nobody fires it by accident.
3. `on: workflow_call:` only for true child workflows, still appears in the sidebar but stays visually quiet since it only runs when called.

jhodapp (changes requested):
@lmcdonough Let me know if you agree with this or not. I'm finding it harder to find the job that I want to run in the list of GH Actions in both repos, so I'd really like this keep this list as short as possible as consolidate the rest as jobs in a few workflow files.

.github/workflows/test-dispatch-regex.yml
lmcdonough: Good catch — addressed in 7b105de. Removed `test-dispatch-regex.yml` and folded the bats suite into `build-test-push.yml` as a parallel `Dispatch Regex Tests` job, so the sidebar stays uncluttered while regex regressions still get caught at PR review time.

Bonus while I was in there: this repo never actually had a `tests/dispatch-pr-preview.bats` file, so the standalone workflow had been a no-op anyway. The new commit adds one with 76 cases — full classifier coverage (PR# / numeric / SHA / branch precedence) plus the new `validate_input` field-protection helper added in 856cdae (length cap, control chars, charset rejection of shell metacharacters / spaces / quotes / parens / etc.). All 76 pass locally.

jhodapp (approved):
Looks awesome @lmcdonough, thanks for seeing this through to an excellent place!

Feel free to merge when ready.

PR #296, opened 2026-04-29
Add MCP server ADR

wumbabum:
## Description
This is an ADR that specs out how an mcp server would look like if added alongside the existing refactor platform rust backend as a new MCP only endpoint.

> [!NOTE]
> The **## Decisions** section of the ADR highlights the main tradeoffs and choices made for the mcp server design, and I welcome any feedback especially to these points, which are all open to debate. The implementation that follows and proof of concept all assume these decisions are set in stone, but could go either way. Likewise for the supporting documents, these are up for debate.

### Changes
* `docs/architecture/mcp_server.md`- Main ADR covering data model (PAT), layer responsibilities, auth, connection protocols, tool design (inputs/outputs/errors), request flow diagram, security considerations, future considerations (OAuth, admin tools, write tools), and implementation decisions.
* `docs/architecture/mcp_server/` - supporting files
  * `mvp_tools.md`- list of tools to implement for the MVP.
  * `exhaustive_list.md`— Full tool inventory. I've dismissed many of these as too risky or low value.
  * `mcp_response_schemas.md`- Request/response schemas for each tool, focusing on structs and shared types.
  * `rmcp_integration.md`- The `rmcp` SDK is a big decision point, this outlines the requirements and benefits of using this dependency. rmcp is pretty standard for mcp servers in the rust ecosystem. There is a decision here to pass self user data via the PAT used during authorization, not a common pattern but not unheard of. The integration takes inspiration from [these mcp servers](https://github.com/modelcontextprotocol/rust-sdk#built-with-rmcp), particularly [apollo-mcp-server](https://github.com/apollographql/apollo-mcp-server) which uses [`RequestContext.extensions`](https://docs.rs/rmcp/latest/rmcp/service/struct.RequestContext.html#structfield.extensions) ([on this line](https://github.com/apollographql/apollo-mcp-server/blob/ff28390b7f3dd18d0f444e14f3eb0c2ffe174e8e/crates/apollo-mcp-server/src/server/states/running.rs#L213)) to feed data to their tools from the initial request. In our case it's the PAT and associated calling user.

### Testing Strategy
This PR is docs-only. However, I plan to use the following testing approach for the implementation. Where possible, tests use existing patterns.
- **Entity API and domain layers** use SeaORM's `MockDatabase` with `#[cfg(feature = "mock")]` gating. Same pattern as existing modules like `note.rs`.
- **Web/middleware tests** use `tower::ServiceExt::oneshot` to send requests through a constructed Axum router with a full mock auth stack. Uses existing `require_auth` approach.
- **MCP tools** are tested by constructing the handler with a mock `AppState`.
- **Integration tests** extend `testing-tools` with an `mcp_client.rs` module that sends JSON-RPC payloads with PAT bearer auth to the live `/mcp` endpoint. This is more or less new to the repo.

AI was used to investigate existing patterns and build on top where possible. The results were stored in a testing-stragegy.md doc and used to validate each step before committing.

### Concerns
1. Read the **Decisions** section in `mcp_server.md` for the key tradeoffs.
2. Review `rmcp_integration.md` for the SDK integration approach — especially the user context propagation pattern discovered in `apollo-mcp-server`.
3. Check `mvp_tools.md` for whether the MVP tool set covers the session summarization use case.

### DIFF
--- a/docs/architecture/mcp_server.md
+++ b/docs/architecture/mcp_server.md
@@ -1,0 +1,169 @@
+# MCP Server
+Add an MCP server to the rust backend so that AI clients can interact with the coaching platform through structured tools.
+
+## Context
+Enable users to interact with the platform without leaving their AI workflow.
+Opens the door to AI automation and generative features like session summarization that the REST API alone cannot provide.
+
+The MCP server will exist inside the existing rust crate as a separate endpoint.
+
+## Data Model
+One new table: `personal_access_tokens`. All MCP tools read from existing tables.
+
+Value is never stored, only the token hash. This means users must be allowed to regenerate their token and deactivate the old one.
+
+Operations: create, show (metadata only, not the raw token), deactivate. No delete — inactive tokens are kept for audit history.
+
+Limit one active PAT per user, enforced by a partial unique index: `UNIQUE (user_id) WHERE status = 'active'`.
+
+| Column | Type | Description |
+|--------|------|-------------|
+| id | uuid PK | |
+| user_id | uuid FK → users | Owner of the token |
+| token_hash | varchar | SHA-256 of the raw token (raw token is never stored) |
+| status | enum `pat_status` (active, inactive) | New Postgres enum, separate from existing `refactor_platform.status`. Only active tokens are accepted. Users can deactivate a lost token and generate a new one. |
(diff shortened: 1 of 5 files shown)


wumbabum:
Closed as the branch name must be lower case

PR #297, opened 2026-04-29
Add MCP server ADR

wumbabum:
## Description
This is an ADR that specs out how an mcp server would look like if added alongside the existing refactor platform rust backend as a new MCP only endpoint.

> [!NOTE]
> The **## Decisions** section of the ADR highlights the main tradeoffs and choices made for the mcp server design, and I welcome any feedback especially to these points, which are all open to debate. The implementation that follows and proof of concept all assume these decisions are set in stone, but could go either way. Likewise for the supporting documents, these are up for debate.

### Proof of concept
#298 is a working mcp server based off this ADR. As part of the POC, I generated local implementation files that call out more detailed technical choices, but ultimately this ADR has the main decision points.

### Changes
* `docs/architecture/mcp_server.md`- Main ADR covering data model (PAT), layer responsibilities, auth, connection protocols, tool design (inputs/outputs/errors), request flow diagram, security considerations, future considerations (OAuth, admin tools, write tools), and implementation decisions.
* `docs/architecture/mcp_server/` - supporting files
  * `mvp_tools.md`- list of tools to implement for the MVP.
  * `exhaustive_list.md`— Full tool inventory. I've dismissed many of these as too risky or low value.
  * `mcp_response_schemas.md`- Request/response schemas for each tool, focusing on structs and shared types.
  * `rmcp_integration.md`- The `rmcp` SDK is a big decision point, this outlines the requirements and benefits of using this dependency. rmcp is pretty standard for mcp servers in the rust ecosystem. There is a decision here to pass self user data via the PAT used during authorization, not a common pattern but not unheard of. The integration takes inspiration from [these mcp servers](https://github.com/modelcontextprotocol/rust-sdk#built-with-rmcp), particularly [apollo-mcp-server](https://github.com/apollographql/apollo-mcp-server) which uses [`RequestContext.extensions`](https://docs.rs/rmcp/latest/rmcp/service/struct.RequestContext.html#structfield.extensions) ([on this line](https://github.com/apollographql/apollo-mcp-server/blob/ff28390b7f3dd18d0f444e14f3eb0c2ffe174e8e/crates/apollo-mcp-server/src/server/states/running.rs#L213)) to feed data to their tools from the initial request. In our case it's the PAT and associated calling user.

### Architecture
Listed in the ADRs, but here's the flow diagram rendered:
Screenshot 2026-05-07 at 7 52 17 PM

### Testing Strategy
This PR is docs-only. However, I plan to use the following testing approach for the implementation. Where possible, tests use existing patterns.
- **Entity API and domain layers** use SeaORM's `MockDatabase` with `#[cfg(feature = "mock")]` gating. Same pattern as existing modules like `note.rs`.
- **Web/middleware tests** use `tower::ServiceExt::oneshot` to send requests through a constructed Axum router with a full mock auth stack. Uses existing `require_auth` approach.
- **MCP tools** are tested by constructing the handler with a mock `AppState`.
- **Integration tests** extend `testing-tools` with an `mcp_client.rs` module that sends JSON-RPC payloads with PAT bearer auth to the live `/mcp` endpoint. This is more or less new to the repo.

AI was used to investigate existing patterns and build on top where possible. The results were stored in a testing-stragegy.md doc and used to validate each step before committing.

### Concerns
1. Read the **Decisions** section in `mcp_server.md` for the key tradeoffs.
2. Review `rmcp_integration.md` for the SDK integration approach — especially the user context propagation pattern discovered in `apollo-mcp-server`.
3. Check `mvp_tools.md` for whether the MVP tool set covers the session summarization use case.


docs/architecture/mcp_server.md
```diff
@@ -0,0 +1,169 @@
...
+- **Data exfiltration**: Tool responses contain names, emails, and session content. This data flows through whatever LLM provider the client uses. Users should be aware their coaching data is sent to third-party models.
+- **Rate limiting**: PAT-scoped rate limits prevent a misbehaving client from enumerating data or overwhelming the backend. Exact limits TBD.
+- **No secrets in responses**: Passwords and PAT values are never included in tool output.
+
+## Future Considerations
+- **OAuth**: PAT is MVP. OAuth enables MCP clients (Claude Desktop, Warp, etc.) to authenticate via browser redirect without manual token generation. The auth middleware will be designed to support a second credential type — OAuth would resolve to the same user + roles as a PAT, reusing the existing authorization layer.
+
+Oauth would require lifecycle management, the mcp server will need to be able to issue access and refresh tokens. Fortunately Axum allows Oauth support through hooks, making it easier to add on later without breaking auth.
+
+- **Admin-only tools**: Post-MVP. Cross-organization data views, user management, platform-wide analytics. Scoped to SuperAdmin role.
```
wumbabum, line 149: One addition that came from discussion with @jhodapp is that even the suggested MVP tools should have admin and super admin authorization rules that are not scoped to coach.

jhodapp: @wumbabum Yes, but only certain things would need this kind of nuance. I'll definitely let you know for any of those cases as they come up while we're working with this. And hopefully much of this nuance is already in place via existing controller call paths, but for any new MCP-specific controllers, we may need to extract user or coaching relationship related context from the client request and route it through a particular `protect` module.

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+10. `create_action` — create action item with optional assignees and goal link
+11. ~~`update_action` — edit action body/due date~~ - editing body text via MCP is awkward; status changes are the high-value operation
+12. ~~`delete_action` — remove action~~ - too destructive for LLM-initiated calls, do this in the UI
+13. `create_note` — add note to a session
+14. ~~`update_note` — edit a note~~ - low value via MCP, notes are typically written once
+15. `create_agreement` — add agreement to a session
+16. ~~`update_agreement` — edit agreement~~ - low value via MCP, agreements are typically written once
+17. ~~`delete_agreement` — remove agreement~~ - too destructive for LLM-initiated calls, do this in the UI
+18. `weekly_digest` — summary across all coachees (generative)
+19. `prepare_for_session` — pre-session brief (generative)
```
jhodapp: I love this!

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+20. ~~`suggest_goals` — suggest goals based on session history (generative)~~ - requires LLM on the server, post-MVP
```
jhodapp, line 7: It might make more sense to start with suggest_actions instead of goals. I'm open to discussing your thoughts around this, however.

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+3. `list_overdue_actions` — overdue actions across all coachees
+4. `create_session` — schedule a session
+5. ~~`update_session` — change date, meeting URL~~ - rescheduling is better done in the UI where you see a calendar
+6. ~~`delete_session` — cancel a session~~ - too destructive for LLM-initiated calls, do this in the UI
+7. `create_goal` — create goal for a coachee
+8. ~~`update_goal` — edit goal title/body~~ - editing body text via MCP is awkward; status changes are the high-value operation
+9. ~~`delete_goal` — remove a goal~~ - too destructive for LLM-initiated calls, do this in the UI
+10. `create_action` — create action item with optional assignees and goal link
+11. ~~`update_action` — edit action body/due date~~ - editing body text via MCP is awkward; status changes are the high-value operation
+12. ~~`delete_action` — remove action~~ - too destructive for LLM-initiated calls, do this in the UI
```
jhodapp: Both a coach and coachee can do these actions through the frontend.

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+13. `create_note` — add note to a session
+14. ~~`update_note` — edit a note~~ - low value via MCP, notes are typically written once
```
jhodapp: This will be interesting, because it's going to need to work through the TipTap REST API to be able to do this. Maybe start with a get_session_notes?

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+15. `create_agreement` — add agreement to a session
+16. ~~`update_agreement` — edit agreement~~ - low value via MCP, agreements are typically written once
+17. ~~`delete_agreement` — remove agreement~~ - too destructive for LLM-initiated calls, do this in the UI
```
jhodapp: Both a coach and coachee can do these actions through the frontend.

docs/architecture/mcp_server/exhaustive_list.md
```diff
@@ -0,0 +1,39 @@
...
+## Shared (both roles, coach specifies coachee, coachee auto-scoped)
+23. `list_sessions` — sessions by date range, sort
+24. `get_session` — full session detail with notes, actions, agreements, goals via `include` param
+25. `list_goals` — goals filterable by status
+26. `list_actions` — actions filterable by session, goal, status, due date
+27. ~~`list_notes` — notes for a session~~ - folds into `get_session` via `include` parameter
+28. ~~`list_agreements` — agreements for a session~~ - folds into `get_session` via `include` parameter
+29. `update_goal_status` — change goal status
+30. `update_action_status` — change action status
+31. `get_session` — session recap (generative)
```
jhodapp:
```suggestion
31. `get_session_recap` — session recap (generative)
```

docs/architecture/mcp_server/rmcp_integration.md
```diff
@@ -0,0 +1,144 @@
...
+
+```rust
+#[derive(Debug, Deserialize, schemars::JsonSchema)]
+struct ListCoacheesParams {
+    // empty — caller identified via PAT in request context
+}
+
+#[tool_router]
+impl McpToolHandler {
+    #[tool(description = "List coachees for the authenticated coach")]
```
jhodapp, line 96: Do the `tool` macros come from another crate (e.g. `rmcp`) or are these intending to be new macros defined here in the backend codebase?

docs/architecture/mcp_server.md
```diff
@@ -0,0 +1,169 @@
...
+Add an MCP server to the rust backend so that AI clients can interact with the coaching platform through structured tools.
+
+## Context
+Enable users to interact with the platform without leaving their AI workflow.
+Opens the door to AI automation and generative features like session summarization that the REST API alone cannot provide.
+
+The MCP server will exist inside the existing rust crate as a separate endpoint.
+
+## Data Model
+One new table: `personal_access_tokens`. All MCP tools read from existing tables.
```
jhodapp, line 11: @calebbourg I'm curious what you think of adding another table for this?

@wumbabum if we eventually switch to using OAuth2 connections, would this new table be needed anymore?

jhodapp (changes requested):
Looking very good, thanks for this, Joe!

docs/architecture/mcp_server.md
wumbabum: It would only not be needed if wanted to drop mcp support for email/password users.

jhodapp: > It would only not be needed if wanted to drop mcp support for email/password users.

That's what I was thinking. I want to think through with @calebbourg as well if it would be just a little more effort to integrate with OAuth2 now and skip this PAT approach, or if the PAT approach as an abstraction on top of the username and password session cookie is a secure and intuitive enough process, even for an MVP.

docs/architecture/mcp_server.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Layer Responsibilities
+
+| Layer | Module | Responsibility |
+|-------|--------|----------------|
+| `entity` | `personal_access_tokens` | SeaORM entity for PAT table |
+| `entity_api` | `personal_access_token` | PAT operations: create, find by token hash, deactivate |
+| `domain` | `personal_access_token` | Business logic: hash on create, validate status |
+| `web` | `mcp/mod` | Module root |
+| `web` | `mcp/auth` | PAT bearer middleware: extract header → hash → lookup user + roles |
```
jhodapp, line 37: If we make sure that the middleware contract in this module is `Bearer <opaque-string> --> Results<users::Model>`, this would ensure that a future OAuth resolver can drop right in alongside of the PAT resolver without having to reshape the tool layer.

docs/architecture/mcp_server/mcp_response_schemas.md
```diff
@@ -0,0 +1,101 @@
+# MCP Request/Response Schemas
+Request and Response types for MCP tool handlers. Entity models already derive `Serialize` with `#[serde(skip_serializing)]` on sensitive fields. Where a tool returns an entity model directly, no wrapper is needed. Where a tool adds computed or nested fields, a thin wrapper uses `#[serde(flatten)]` to inline the entity and only defines the extra fields.
+
+## `get_coachee`
+
+### Input
+
+| Param | Type | Required | Description |
+|-------|------|----------|-------------|
+| coachee_id | uuid? | Coaches: yes. Coachees: no (defaults to self). | The coachee to look up. |
```
calebbourg, line 10: Users can be both a coach and coachee at the same time. Ex. Joseph is Jim's coach and Jim is my coach. Jim is both a coach and Coachee.

Just calling this out in case we need to think about including that context to identify a tool call as "This is from user A acting as a Coach" vs "This is user A acting as a Coachee"

docs/architecture/mcp_server/rmcp_integration.md
```diff
@@ -0,0 +1,144 @@
...
+
+This eliminates the need for `Arc<RwLock<...>>` bridging or storing the user on the handler struct. The `McpToolHandler` only needs `app_state` — the user flows through the request context on every call.
+
+## Stateless mode (MVP)
+
+[`StreamableHttpServerConfig`](https://docs.rs/rmcp/latest/rmcp/transport/streamable_http_server/tower/struct.StreamableHttpServerConfig.html) `{ stateful_mode: false, .. }` — every POST is self-contained. No session ID, no persistent handler. The factory creates a fresh `McpToolHandler` per request.
+
+Stateful mode (`stateful_mode: true`, the default) creates persistent sessions where the handler lives across multiple requests. Deferred to post-MVP for SSE streaming.
```
calebbourg, line 66: Curious what sort of use cases something like this would unlock

docs/architecture/mcp_server.md
```diff
@@ -0,0 +1,169 @@
...
+| `web` | `mcp/tools/mod` | `McpToolHandler` struct + `#[tool_router]`/`#[tool_handler]` wiring. Holds `AppState`. |
+| `web` | `mcp/tools/*` | Individual tool methods using `#[tool]` macro |
+| `web` | `router` | MCP service nested via `nest_service("/mcp", service)` in existing `define_routes()` |
+| `web` | `controller/pat_controller` | REST endpoints for PAT management (create, show, deactivate) used by the UI |
+| `web` | `protect/users/tokens` | Authorization middleware for PAT routes: user can only manage own tokens |
```
calebbourg: We're in the process of moving away from this model and using extractors directly for authorization. 

Here's a PR that @rnambaale put up that shows the new pattern https://github.com/refactor-group/refactor-platform-rs/pull/239

calebbourg (commented):
@wumbabum This is really great and well thought out. Awesome work!

I left a couple of small comments. 

I'm on board with using PATs for the first iteration of this and tying it into a given user's pre-existing auth permissions.

docs/architecture/mcp_server/exhaustive_list.md
wumbabum: Good thinking, goals are often the result of human collaboration through the UI, so less useful for updating using the mcp server. But actions are great to brainstorm via an agent session.

docs/architecture/mcp_server/exhaustive_list.md
wumbabum: Agreed, this is not going to be a part of MVP or probably even post MVP.

docs/architecture/mcp_server.md
wumbabum: > skip this PAT approach
There's an advantage to having a PAT setup, and it's that the codeowners control every part of that process.

I've found Oauth to be somewhat less reliable for setting up mcp usage. The issue is that configurations look different depending on what AI tool you're using (Claude, warp). It also depends on having a reliable, unchanging oidc process, which the the IdP (google) controls. by comparison the api for the PAT is unlikely to ever need to change.

Of course, the downside is it takes work to support both.

docs/architecture/mcp_server/rmcp_integration.md
wumbabum: These macros come from `rmcp`, and help defined required features of [mcp](https://modelcontextprotocol.io/specification/2025-11-25#features) protocol

docs/architecture/mcp_server.md
wumbabum: Yes exactly, both auth models should resolve to shared users model, and reuse as much as possible of the existing authorization logic. Great point.

docs/architecture/mcp_server/mcp_response_schemas.md
wumbabum: Hm, you're right. This is making me think coachee_id should be required. How would the caller being a coach vs the coachee affect the authorization or the response?

I'm assuming the caller is authorized if they're the coachee themself, or they're a coach for that coachee.
I'm assuming response wouldn't change either way. Unless there's PII?

docs/architecture/mcp_server/rmcp_integration.md
wumbabum: Yeah honestly can't think of any right now. Maybe something where a live session is interacting with a user's ai session? Definitely a v2 type of feature.

docs/architecture/mcp_server.md
jhodapp: @wumbabum So just to get this recorded here, we agreed on Friday that you would start here by finishing the migration away from the protect modules and to using extractors directly for authorization. That'll make a solid and less confusing foundation on which to build any new kinds of authorization rule checks you may want to add when implementing your MCP server ADR.

docs/architecture/mcp_server.md
jhodapp: Or per our plan in the last comment, not through a protect module, but through an auth extractor once the full port away from protect modules is complete.

docs/architecture/mcp_server.md
```diff
@@ -0,0 +1,171 @@
...
+|--------|------|-------------|
+| id | uuid PK | |
+| user_id | uuid FK → users | Owner of the token |
+| token_hash | varchar | SHA-256 of the raw token (raw token is never stored) |
+| status | enum `pat_status` (active, inactive) | New Postgres enum, separate from existing `refactor_platform.status`. Only active tokens are accepted. Users can deactivate a lost token and generate a new one. |
+| last_used_at | timestamptz | Optional, for observability |
+| created_at | timestamptz | |
+| updated_at | timestamptz | |
+
+## Layer Responsibilities
```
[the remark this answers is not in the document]
jhodapp: @wumbabum worth noting here.

jhodapp (approved):
@wumbabum Looks great, thanks for this effort! I think it's ready to begin the step-by-step implementation we discussed on Friday, starting with finishing the migration from protect modules to extractors for the entire backend.

Then moving on to implementing the PAT handling with a paired frontend branch for creating and viewing the PAT from the user's Settings page.

Then taking this a command or two at a time so we can get some experience on it especially around the security concerns before going into production with all proposed commands.

PR #301, opened 2026-05-08
Bring backend env to functional parity with prod

jhodapp:
## Description

PR preview deploys were omitting OAuth, encryption, and most email-notification env vars that production injects into the backend container. Anything in `domain/oauth_connection.rs` (which reads `encryption_key`, `google_client_id`, `google_client_secret`, `google_redirect_uri` via `Config`) returned a configuration error in PR previews, and session-scheduled / action-assigned email links couldn't be constructed without `FRONTEND_BASE_URL`.

This brings PR preview to **functional parity** with prod — only adding env vars that the Rust `Config` struct (`service/src/config.rs`) actually reads at runtime AND that have no correct default already baked into the struct. Variables prod sets that are inert to the running app, or that already have correct Clap `default_value` declarations, are deliberately excluded.

### Changes

* Added the OAuth credentials cluster (`ENCRYPTION_KEY`, `GOOGLE_CLIENT_ID`/`SECRET`/`REDIRECT_URI`) and the per-PR `OAUTH_SUCCESS_REDIRECT_URI` to the env-file generator in `ci-deploy-pr-preview.yml` and the backend service block in `docker-compose.pr-preview.yaml`.
* Added the missing email cluster: `SESSION_SCHEDULED_*` and `ACTION_ASSIGNED_*` template IDs, and `FRONTEND_BASE_URL` (constructed dynamically from `RPI5_TAILSCALE_NAME` + PR number).
* OAuth secrets default to `'UNUSED'` when corresponding `PR_PREVIEW_*` secrets are not configured, matching the existing `TIPTAP_*` / `MAILERSEND_*` fallback pattern. Existing PR preview deploys are unaffected; OAuth is opt-in once secrets are populated.

### Deliberately excluded

**No-op vars** (verified not consumed by Rust at runtime):
* Discrete `POSTGRES_USER` / `PASSWORD` / `DB` / `HOST` / `PORT` / `OPTIONS` / `SSL_ROOT_CERT`. Prod uses these to construct `DATABASE_URL` for DO Managed Postgres; PR preview's `DATABASE_URL` is a fixed string targeting the in-cluster `postgres` container.
* `BACKEND_API_VERSION` — Rust reads `API_VERSION` (Clap derives env name from field `api_version`).
* `PLATFORM` and `BACKEND_IMAGE_NAME` — docker-compose-time only, never read by the running app.

**Vars with correct Rust defaults** (drift hazard if duplicated; one of these already had a divergent fallback bug in commit 1 of this PR):
* `MAILERSEND_BASE_URL`, `SESSION_SCHEDULED_EMAIL_URL_PATH`, `ACTION_ASSIGNED_EMAIL_URL_PATH`
* `GOOGLE_OAUTH_AUTH_URL`, `GOOGLE_OAUTH_TOKEN_URL`, `GOOGLE_USERINFO_URL`, `GOOGLE_MEET_API_URL`

For each of these, `service/src/config.rs` already pins the canonical value via `#[arg(env, default_value = "…")]`. Setting the env var again at deploy time would either duplicate the value (extra place to update) or, if a future fallback drifts, silently override the correct in-code default.

### Testing Strategy

1. Manually dispatch `dispatch-pr-preview.yml` targeting any open PR. Confirm the stack deploys green and `docker compose -p pr-<NUM> -f ~/pr-<NUM>-compose.yaml ps` shows backend and frontend healthy.
2. `docker compose -p pr-<NUM> -f ~/pr-<NUM>-compose.yaml exec backend env | grep -E '^(ENCRYPTION|GOOGLE_CLIENT|GOOGLE_REDIRECT|FRONTEND_BASE|SESSION_SCHEDULED|ACTION_ASSIGNED|OAUTH_SUCCESS)'` to confirm the new vars are populated (most should be `UNUSED` until corresponding secrets are added).
3. (Optional) Add real values for `PR_PREVIEW_ENCRYPTION_KEY`, `PR_PREVIEW_SESSION_SCHEDULED_EMAIL_TEMPLATE_ID`, etc., and confirm the backend container env reflects them on next deploy.

### Concerns

**OAuth flow does NOT yet function end-to-end in PR previews.** The wildcard parent redirect strategy this PR is built around — one URI registered in the Google OAuth console, backend bounces to `/pr-<NUM>/` via OAuth `state` param after token exchange — requires follow-on changes:

* New backend route in `meeting-auth/` (or `domain/src/oauth_connection.rs`) handling the shared `/oauth/google/preview-callback` path and parsing the PR number out of the OAuth `state` parameter.
* New nginx-preview route to forward the shared callback path to a designated backend container.

Until those land, leaving OAuth secrets at the `'UNUSED'` default keeps existing PR preview deploys functioning identically to today. The compose env block contains an inline `NOTE` referencing this.

No backend code changes in this PR; no migrations; no breaking changes for callers.


jhodapp:
Tracking the prefix-divergence cleanup separately in #302 — the new `PR_PREVIEW_*` references this PR adds follow the existing convention in [ci-deploy-pr-preview.yml](.github/workflows/ci-deploy-pr-preview.yml) to keep the diff scoped to env parity. The full rename across both repos' `pr-preview` environments will land in a follow-on PR coordinated with refactor-platform-fe.

PR #308, opened 2026-05-11
Implements recurring coaching sessions

rnambaale:
## Description

Adds support for scheduling a recurring series of coaching sessions in one request via `POST /coaching_sessions/recurring`, instead of a coach having to create each weekly (or bi-weekly / monthly / daily) session by hand.

Because the existing one-off create path eagerly creates a Tiptap collab document, mints a meeting URL via the coach's OAuth provider, and links in-progress goals — doing all of that up front for a 24-session series would fan out to ~50 external API calls and waste resources nobody may ever use. So the recurring endpoint persists only the row state and **defers** those three side effects to **first read** of each session. A Postgres advisory lock keyed on the session id serializes concurrent first-load requests so the work happens exactly once per session even if coach and coachee open it at the same instant.

### Changes

* New `POST /coaching_sessions/recurring` endpoint. Coach-only. Accepts a `Recurrence` rule (frequency: daily / weekly / biweekly / monthly, optional `interval`, optional `by_weekdays` for weekly multi-day patterns, and one of `count` or `until`). Capped at 365 occurrences and 366 calendar days regardless of frequency.
* `hydrated_at TIMESTAMPTZ` column on `coaching_sessions` flagging whether the row's deferred side effects have run. Existing rows are backfilled with `created_at` so the read path's hydration check skips them. Eager one-off create stamps `hydrated_at = NOW()` inline.
* `domain::coaching_session::ensure_hydrated` runs the deferred side effects (Tiptap doc create, provider + meeting URL resolution via `oauth_connection::find_by_user`, in-progress goal linking) and stamps `hydrated_at`, serialized by a per-session `pg_advisory_xact_lock`. Wired into `GET /coaching_sessions/:id` so first-read populates the row.
* In-progress goal auto-link switched from N per-row inserts to a single `INSERT ... ON CONFLICT (coaching_session_id, goal_id) DO NOTHING` so concurrent or replay hydration is safe and the eager create path stays correct.
* New `notify_recurring_sessions_scheduled` email orchestrator sends a single series-summary email per recipient (coach, coachee), rather than one-per-session which would spam for long series. Mirrors the one-off `notify_session_scheduled` shape.
* New `recurring_sessions_scheduled_email_template_id` config field. **The MailerSend template itself needs to be created out-of-band** with the matching ID; without it, the orchestrator logs a config error and the request still succeeds.

#### Automated

* `cargo test --workspace` — adds unit + mock-DB tests covering: recurrence rule validation and expansion (21 cases in `recurrence::tests` — daily, weekly, weekly+by_weekdays, biweekly, monthly day-of-month clamping, cap rejection); idempotent goal-link replay; `bulk_create_recurring` happy path + truncation; `ensure_hydrated` short-circuit on already-hydrated input; `mark_hydrated` UPDATE shape; `oauth_connection::find_by_user` for the zero / one / multiple-rows cases.
* `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean.

#### Manual E2E

1. As a coach with an active Zoom OAuth connection: `POST /coaching_sessions/recurring` with `frequency=weekly`, `by_weekdays=["Mon","Wed"]`, `count=4`. Expect 8 rows returned with `provider=null`, `meeting_url=null`, `collab_document_name=null`, `hydrated_at=null`.
2. `GET /coaching_sessions/:id` for one of them. Expect the response to now have `provider="zoom"`, populated `meeting_url`, populated `collab_document_name`, non-null `hydrated_at`. Verify in Tiptap dashboard that the doc exists, and in Zoom that the meeting was created.
3. Re-`GET` the same session. Response is identical to step 2; no new Tiptap / Zoom resources created.
4. As the coachee, `GET` a sibling unhydrated session. Same hydration outcome.
5. **No-OAuth path**: as a coach with no OAuth connection, `POST` a series and then `GET` one. `provider` and `meeting_url` stay `null`, but `collab_document_name`, `hydrated_at`, and goal links are populated.
6. **Goal auto-link**: create an in-progress goal on the relationship, then `GET` an unhydrated session → goal is now linked. Repeat the `GET`; no duplicate-link error.

### Concerns

* **MailerSend template** for the series-summary email isn't created yet — needs to be set up against `RECURRING_SESSIONS_SCHEDULED_EMAIL_TEMPLATE_ID` in prod and PR-preview env. Until that's done, recurring-create requests still succeed but no email goes out.


domain/src/emails.rs
```diff
@@ -453,6 +466,145 @@ pub async fn notify_session_scheduled(
...
+
+    let email_request = SendEmailRequestBuilder::new()
+        .from("hello@myrefactor.com")
+        .to_with_name(
+            &recipient.email,
+            format!("{} {}", recipient.first_name, recipient.last_name),
+        )
+        .subject(format!(
+            "{session_count} recurring coaching sessions scheduled"
+        ))
```
jhodapp: Grammar: this produces "1 recurring coaching sessions scheduled" when
`session_count` is 1. Should use the singular form for a count of one.

Suggested fix:

```rust
    .subject(if sessions.len() == 1 {
        "1 recurring coaching session scheduled".to_string()
    } else {
        format!("{session_count} recurring coaching sessions scheduled")
    })
```

The test on line 1443 asserts the incorrect string and will also need updating.

entity_api/src/coaching_session.rs
```diff
@@ -69,6 +109,51 @@ pub async fn delete(db: &impl ConnectionTrait, coaching_session_id: Id) -> Resul
...
+    .await?;
+    Ok(())
+}
+
+fn advisory_lock_key(session_id: Id) -> i64 {
+    let bytes = session_id.as_bytes();
+    let mut upper = [0u8; 8];
+    upper.copy_from_slice(&bytes[..8]);
+    i64::from_be_bytes(upper)
+}
```
jhodapp, line 135: This method discards the lower 8 bytes of the UUID entirely, giving
up half the entropy. Two sessions whose UUIDs share the same upper 8 bytes
map to the same advisory lock key, causing unnecessary serialization. Low
probability with UUID v4, but avoidable by folding all 16 bytes:

```rust
    fn advisory_lock_key(session_id: Id) -> i64 {
        let bytes = session_id.as_bytes();
        let hi = i64::from_be_bytes(bytes[..8].try_into().unwrap());
        let lo = i64::from_be_bytes(bytes[8..].try_into().unwrap());
        hi ^ lo
    }
```

web/src/controller/coaching_session_controller.rs
```diff
@@ -141,6 +154,61 @@ pub async fn create(
...
+
+    let db = app_state.db_conn_ref();
+
+    let relationship =
+        CoachingRelationshipApi::find_by_id(db, params.coaching_relationship_id).await?;
+    if relationship.coach_id != user.id {
+        return Err(Error::Web(WebErrorKind::Auth));
+    }
+
+    let dates = CoachingSessionApi::expand_recurrence(params.start_at, &params.recurrence)
```
jhodapp: The auth check runs after a DB round-trip to fetch the relationship. Consider
flipping the order: validate the recurrence rule first (pure, no I/O), then
fetch the relationship and check coach ownership. Invalid recurrence rules
currently hit the DB unnecessarily before failing.

```rust
    let dates = CoachingSessionApi::expand_recurrence(params.start_at, &params.recurrence)
        .map_err(|e| ...)?;

    let relationship =
        CoachingRelationshipApi::find_by_id(db, params.coaching_relationship_id).await?;
    if relationship.coach_id != user.id {
        return Err(Error::Web(WebErrorKind::Auth));
    }
```

entity_api/src/coaching_session_goal.rs
```diff
@@ -317,11 +321,27 @@ pub async fn link_in_progress_goals_to_session(
...
+        .do_nothing()
+        .to_owned();
+
+    let inserted = Entity::insert_many(rows)
+        .on_conflict(on_conflict)
+        .exec_with_returning_many(db)
+        .await?;
 
-    Ok(in_progress_goals.len())
+    Ok(inserted.len())
```
jhodapp, line 344: Heads-up semantic change: this function used to return the total number of
in-progress goals found (even if all were already linked). Now it returns the
count of *newly inserted* rows (zero if every goal was already linked due to
ON CONFLICT DO NOTHING). Any callers that treated the return value as "total
goals" rather than "new links added" will now silently see 0 on replay. The
doc comment on line 303 already reflects the new meaning, but worth a callout
since the signature is unchanged.

domain/src/coaching_session/recurrence.rs
```diff
@@ -0,0 +1,688 @@
...
+    let monday_of_start_week =
+        start_at.date() - Duration::days(start_at.weekday().num_days_from_monday() as i64);
+
+    let mut out = Vec::new();
+    let limit = target_count.unwrap_or(gen_cap).min(gen_cap);
+    let mut week_offset: i64 = 0;
+    // Safety cap on outer loop iterations to bound worst-case work even when
+    // `until` is far in the future and weekday filters mean few hits per week.
+    let max_weeks = (gen_cap as i64) + 8;
+    while out.len() < limit && week_offset < max_weeks * week_interval as i64 {
```
jhodapp: Nit: operator precedence here is surprising — `as` binds tighter than `*`,
so this reads as `max_weeks * (week_interval as i64)`, which is correct, but
a reader could easily misread it. Add parens to make the intent explicit:

```rust
    while out.len() < limit && week_offset < max_weeks * (week_interval as i64) {
```

domain/src/coaching_session.rs
```diff
@@ -1,3 +1,5 @@
+pub mod recurrence;
```
jhodapp: `pub mod recurrence` makes the module itself navigable from outside the crate
(`domain::coaching_session::recurrence::*`). Since everything callers need is
already re-exported at the `coaching_session` level (lines 21-24), the module
path is unnecessary public API surface. Change to `pub(crate)`:

```rust
    pub(crate) mod recurrence;
```

domain/src/coaching_session.rs
```diff
@@ -16,6 +18,10 @@ use service::config::Config;
 pub use entity_api::coaching_session::{
     find_by_id, find_by_user_with_includes, EnrichedSession, IncludeOptions, SessionQueryOptions,
 };
+pub use recurrence::{
+    expand_recurrence, validate_recurrence, Frequency, Recurrence, RecurrenceError,
+    MAX_RECURRING_OCCURRENCES, MAX_RECURRING_SPAN_DAYS,
+};
```
jhodapp: `validate_recurrence` is only called internally by `expand_recurrence` — it has
no callers in the web layer and its tests live inside the recurrence module's
own `#[cfg(test)]` block. Exporting it as public API allows callers to invoke
it out of sequence (validate with one set of args, expand with another).
Remove it from this re-export list; `expand_recurrence` is the public contract.

web/src/controller/coaching_session_controller.rs
```diff
@@ -141,6 +154,61 @@ pub async fn create(
...
+        .map_err(|e| DomainError {
+            source: None,
+            error_kind: DomainErrorKind::Validation(e.message()),
+        })?;
```
jhodapp: The web layer is constructing a `DomainError` struct directly, which violates
the layer contract in coding-standards.md: "Errors must flow through the layer
chain entity_api → domain → web without skipping layers."

The `RecurrenceError → domain::Error` conversion belongs in the domain layer as
a `From` impl:

```rust
    // domain/src/coaching_session/recurrence.rs
    impl From<RecurrenceError> for domain::Error {
        fn from(e: RecurrenceError) -> Self {
            domain::Error {
                source: None,
                error_kind: DomainErrorKind::Validation(e.message()),
            }
        }
    }
```

With that in place, `expand_recurrence` can return `Result<Vec<NaiveDateTime>,
domain::Error>` and the controller just uses `?` with no manual mapping or
domain type imports needed.

domain/src/coaching_session.rs
```diff
@@ -107,6 +114,94 @@ pub async fn create(
...
+
+    let mut session = coaching_session::find_by_id(&txn, session_id).await?;
+    if session.hydrated_at.is_some() {
+        txn.commit().await.map_err(entity_api::error::Error::from)?;
+        return Ok(session);
+    }
+
+    let coaching_relationship =
+        coaching_relationship::find_by_id(db, session.coaching_relationship_id).await?;
+    let organization = organization::find_by_id(db, coaching_relationship.organization_id).await?;
```
jhodapp, line 164: Minor style inconsistency with the existing `create` function: the transaction
(`txn`) is open and holding the advisory lock at this point, but the
relationship and organization reads bypass it via the raw `db` handle. `create`
uses the transaction consistently for all operations after it begins.

The reads don't need to be inside the transaction for correctness here, but a
future reader will reasonably wonder whether the bypass is intentional. A short
comment noting that these are read-only lookups that don't need the transaction
scope would remove the ambiguity.

web/src/controller/coaching_session_controller.rs
```diff
@@ -141,6 +154,61 @@ pub async fn create(
...
+
+    let sessions =
+        CoachingSessionApi::bulk_create_recurring(db, params.coaching_relationship_id, dates)
```
jhodapp, line 197: The relationship is fetched twice: once here to check `coach_id` (line 192),
then again inside `bulk_create_recurring` to validate it exists
(domain/src/coaching_session.rs:125). Both calls hit the same row.

One option: pass `user.id` into `bulk_create_recurring` and let it do the
single fetch, check existence and coach ownership together, and return an auth
error if the caller isn't the coach. The controller then has one fewer DB
round-trip and one fewer import (`CoachingRelationshipApi`).

jhodapp (changes requested):
Thanks Raymond, this looks very good. Just a few things to address from my POV:

1. Can you provide me instructions for what I need to do to update the email that gets sent out for a scheduled reoccurring sessions? Were you intending on this being a new email separate from the existing single scheduled session or combine into the same one with some logic?

Here's the current email template as it stands right now:

```
Hi, {{first_name}}.

A new coaching session has been scheduled.


You have an upcoming session with your {{other_user_role}}, {{other_user_first_name}} {{other_user_last_name}}, on:

{{session_date}} at {{session_time}}

[View Session](https://myrefactor.com/%7B%7Bsession_url%7D%7D)

If you have any questions, reach out to your {{other_user_role}} directly.

— The {{organization_name}} Team
```

2. Can you also be sure to update the [DBML](docs/db/refactor_platform_rs.dbml) and [SVG diagram](docs/db/refactor_platform_rs.png).

domain/src/emails.rs
rnambaale: Fixed: 961d7984fda2d45e87b600c12c9cbd6446e54fa4

entity_api/src/coaching_session.rs
rnambaale: Fixed: e5580432b95e1d2894b9ca05dc86b15d6d8111bd

web/src/controller/coaching_session_controller.rs
rnambaale: Fixed

entity_api/src/coaching_session_goal.rs
rnambaale: For the case of recurring sessions where where goals are not linked at creation, there is a chance some of in-progress goals would have been linked manually by hydration time.

I think it reads more accurately if this returned the goals that were actually linked by the function call. Happy to iterate further on this.

domain/src/coaching_session/recurrence.rs
rnambaale: Fixed

domain/src/coaching_session.rs
rnambaale: I have resolved this. I actually made `validate_recurrence` private. 

7bcc0136448a413d5565de9a2ef109048db50c64

web/src/controller/coaching_session_controller.rs
rnambaale: Fixed

web/src/controller/coaching_session_controller.rs
rnambaale: In my opinion, this is an AuthZ check that belongs to the controller, so taking it to the domain layer would contradict the existing patterns.

Since `domain::coaching_sessions::bulk_create_recurring` is internal, the application knows to call it only when the relationship is a valid, so I have removed the redundant check in the domain. 49da83efccc920016d2f187c62298c7cb1396d85

Happy to revisit this if you think there is a better alternative.

rnambaale:
> 1. Can you provide me instructions for what I need to do to update the email that gets sent out for a scheduled reoccurring sessions? Were you intending on this being a new email separate from the existing single scheduled session or combine into the same one with some logic?

@jhodapp 

So what I have in mind is for this to be a new email separate from the exsiting one that gets sent out when a single session is created.

It would read something like:

```
Hi, {{first_name}}.

{{session_count}} new coaching sessions have been scheduled.


You have a recurring series of sessions with your {{other_user_role}}, {{other_user_first_name}} {{other_user_last_name}}.

The series begins on {{first_session_date}} at {{first_session_time}} and runs through {{last_session_date}}.

[View First Session](https://myrefactor.com/%7B%7Bsession_url%7D%7D)

If you have any questions, reach out to your {{other_user_role}} directly.

— The {{organization_name}} Team

```

rnambaale:
> 2\. Can you also be sure to update the [DBML](docs/db/refactor_platform_rs.dbml) and [SVG diagram](docs/db/refactor_platform_rs.png).

DB schema docs have been updated. c11fef2899a99bc1f402d22fb2783946f39e9fd5

jhodapp:
@rnambaale thanks for this, I'll get these changes tested locally here soon and get back to you with feedback or approval.

FYI, I just landed a switch from MailerSend to Resend and that's resulted in a merge conflict on your branch with main.

I'll get your new email template added to Resend as well, thanks for providing it.

domain/src/emails.rs
```diff
@@ -493,27 +509,22 @@ async fn send_recurring_series_email_to_recipient(
...
+        .add_variable("other_user_first_name", &other_user.first_name)
+        .add_variable("other_user_last_name", &other_user.last_name)
+        .add_variable("other_user_role", other_user_role)
+        .add_variable("organization_name", &organization.name)
+        .add_variable("session_count", &session_count)
+        .add_variable("first_session_date", &first_session_date)
+        .add_variable("first_session_time", &first_session_time)
+        .add_variable("last_session_date", &last_session_date)
+        .add_variable("session_url", &session_url)
         .build()
```
rnambaale: The switch to resend expects the email subject to be computed within the template.

jhodapp: @rnambaale actually the opposite, I set the title from the Resend template editor itself.

rnambaale: Right, I think for this case the email subject is not static. I have added the option to customise the subject from the app. Is that inline with what you were expecting?

jhodapp: @rnambaale No, you can just pass the needed template variable(s) and I can use it in the subject line of the template builder in Resend. So you don't need to set the .subject key at all, just leave it unset.

entity_api/src/coaching_session_goal.rs
jhodapp: Oh yes, we do want it reflecting reality. So if what you mentioned is what happens, then I agree with you since this is different than creating a single new coaching session where the goals automatically get linked to the new session at creation-time.

web/src/controller/coaching_session_controller.rs
jhodapp: This works well, thanks!

jhodapp:
> > 1. Can you provide me instructions for what I need to do to update the email that gets sent out for a scheduled reoccurring sessions? Were you intending on this being a new email separate from the existing single scheduled session or combine into the same one with some logic?
> 
> @jhodapp
> 
> So what I have in mind is for this to be a new email separate from the exsiting one that gets sent out when a single session is created.
> 
> It would read something like:
> 
> ```
> Hi, {{first_name}}.
> 
> {{session_count}} new coaching sessions have been scheduled.
> 
> 
> You have a recurring series of sessions with your {{other_user_role}}, {{other_user_first_name}} {{other_user_last_name}}.
> 
> The series begins on {{first_session_date}} at {{first_session_time}} and runs through {{last_session_date}}.
> 
> [View First Session](https://myrefactor.com/%7B%7Bsession_url%7D%7D)
> 
> If you have any questions, reach out to your {{other_user_role}} directly.
> 
> — The {{organization_name}} Team
> ```

Great thanks, I'll get this template in place before we merge and deploy this.

jhodapp (approved):
Looks great, thanks for this fine work!

PR #312, opened 2026-05-14
Migrate transactional email from MailerSend to Resend

jhodapp:
## What this does

Swaps our transactional email provider from MailerSend to Resend. MailerSend gates the number of templates by pricing tier, and we need a 4th template (password reset) — Resend's pricing is volume-based with no per-template gating, so the switch removes that ceiling. We'll be able to get what we need for quite a while on their free tier to start, and then it's a better price jump to \$20/month for sending up to 50k emails per month. MailerSend jumps to \$35/month just to get a 4th template, a deal-breaker.

The change is deliberately low-risk. Our MailerSend integration was a thin `reqwest` wrapper, not a vendor SDK, so this is mostly a payload-shape rewrite in one gateway file plus config and env-var renames — controllers and the `domain::emails` orchestration are untouched. Resend's API uses the same `template id + variables` model we already had, so it's close to a drop-in.

Test coverage was reworked alongside the swap: the gateway now has direct HTTP-behavior tests, and the integration tests body-match per recipient, so a per-recipient bug (a role swap, a mixed-up assignee) actually fails the suite.

## Sending domain

All emails go out as `hello@mail.myrefactor.com`. Using a dedicated subdomain rather than the root domain keeps the Resend DNS records (SPF, DKIM) in their own zone and isolates sending reputation — if the root domain ever gains its own email usage (human mailboxes, MX records), its SPF record stays clean.

## Before merging / deploying

This PR is the code half of the migration. The steps below need a Resend account and GitHub repo admin, and **must be done before deploy** — otherwise email silently breaks:

1. **Verify `mail.myrefactor.com` in Resend** (SPF + DKIM DNS records). The sending address is hardcoded to `hello@mail.myrefactor.com` in this PR.
2. **Author the 3 templates** in the Resend dashboard — welcome, session-scheduled, action-assigned — using the variable names the backend sends. (The 4th template, password-reset, belongs to its own branch.) The exact per-template variable list is in `docs/implementation-plans/email-notifications-resend.md`.
3. **Add the GitHub Secret** `RESEND_API_KEY` (production) and `PR_PREVIEW_RESEND_API_KEY` (pr-preview environment). The old `MAILERSEND_API_KEY` secrets can stay until Resend is confirmed working in prod — rollback insurance.
4. **Update the three template-ID Variables** (`WELCOME_EMAIL_TEMPLATE_ID`, `SESSION_SCHEDULED_EMAIL_TEMPLATE_ID`, `ACTION_ASSIGNED_EMAIL_TEMPLATE_ID`) to the new Resend template IDs. These already exist as GitHub Variables; the names stay the same, only the values change.

A detailed operational runbook (domain verification, DNS record types, which GitHub Variables vs Secrets to touch, deploy + verify steps) is available separately.

## Note

Resend reserves the variable names `FIRST_NAME`, `LAST_NAME`, `EMAIL`, `UNSUBSCRIBE_URL`. Ours are lowercase (`first_name`, etc.) and should be fine, but worth confirming on the first real send.

## Test plan

Once the steps above are done:

- [ ] `dig TXT mail.myrefactor.com` shows SPF + DKIM; Resend reports domain Verified
- [ ] Welcome email: create a user → email arrives at `hello@mail.myrefactor.com` with variables rendered
- [ ] Resend-invite works
- [ ] Session scheduled: both coach and coachee receive the email, dates shown in their own timezones
- [ ] Action assigned with multiple assignees: each gets a distinct, correctly-addressed email
- [ ] Resend dashboard Logs show each send with an `id` matching the backend log line
- [ ] PR preview environment deploys with the renamed secret


domain/src/emails.rs
```diff
@@ -154,19 +154,19 @@ async fn send_welcome_email(
         .map(|b| b.build(TOKEN_PLACEHOLDER, magic_link_token))
         .unwrap_or_default();
 
-    debug!("Preparing personalization data for {}", user.email);
+    debug!("Preparing template variables for {}", user.email);
 
     let email_request = SendEmailRequestBuilder::new()
-        .from("hello@myrefactor.com")
+        .from("hello@mail.myrefactor.com")
```
rnambaale: I think we can consider defining this as a constant.

rnambaale (approved):
This looks good! Just one thought.

domain/src/emails.rs
jhodapp: @rnambaale yes indeed, good catch!

jhodapp: Fixed in be0673c.

PR #313, opened 2026-05-17
Unify PR preview env var names with production; reclassify non-sensitive items as vars

jhodapp:
## Description

Drops the `PR_PREVIEW_*` prefix from all references in `ci-deploy-pr-preview.yml` so secret/var names match production's `deploy_to_do.yml`. Also reclassifies non-sensitive items (DB name, schema, ports, hostnames, template IDs, API version strings, OAuth client ID, redirect URI) from `secrets.*` to `vars.*`, matching production's split.

Resolves the workflow's own header TODO (line 13 pre-change): *"Move PR_PREVIEW_* secrets to org-level for proper cross-repo access."* The prefix existed to avoid collisions when GitHub Actions environments didn't isolate secrets cleanly; environments now isolate them natively, so the namespace prefix is no longer load-bearing.

Coordinated cross-repo via the coordinator board under question `pr_preview_secrets_rename_and_reclassify`. Frontend confirmed:
- Zero direct `PR_PREVIEW_*` references in any FE workflow file
- The FE repo's `pr-preview` GitHub environment holds its own copies of the same names; FE will rename in lockstep
- They chose Option B (full coordinated cleanup, not incremental)

### Changes

- **Rename** every `PR_PREVIEW_*` reference to its unprefixed equivalent (~25 sites)
- **Reclassify** items from `secrets.*` → `vars.*` to match production's split:
  - `POSTGRES_DB`, `POSTGRES_SCHEMA`
  - `TIPTAP_APP_ID`
  - `WELCOME_EMAIL_TEMPLATE_ID`, `SESSION_SCHEDULED_EMAIL_TEMPLATE_ID`, `ACTION_ASSIGNED_EMAIL_TEMPLATE_ID`
  - `BACKEND_SERVICE_PROTOCOL`, `BACKEND_SERVICE_HOST`, `BACKEND_SERVICE_PORT`, `BACKEND_SERVICE_API_PATH`, `BACKEND_API_VERSION`
  - `FRONTEND_SERVICE_PORT`, `FRONTEND_SERVICE_INTERFACE`
  - `GOOGLE_CLIENT_ID`, `GOOGLE_REDIRECT_URI`
- **Keep as secrets** (rename only): `POSTGRES_USER`, `POSTGRES_PASSWORD`, `TIPTAP_URL`, `TIPTAP_AUTH_KEY`, `TIPTAP_JWT_SIGNING_KEY`, `RESEND_API_KEY`, `ENCRYPTION_KEY`, `GOOGLE_CLIENT_SECRET`
- **Declare two previously-undeclared secrets** (`ENCRYPTION_KEY`, `GOOGLE_CLIENT_SECRET`) in the `secrets:` input block for consistency
- **Remove three resolved TODO comments** (header at line 13, plus two inline duplicates)

Net diff: 46 insertions / 72 deletions.

### Testing Strategy

**Required before merge — cross-repo smoke test:**

1. Coordinate with FE to rename their `pr-preview` environment secrets/vars in lockstep with this PR. Same set of names; their environment in the FE repo needs to mirror this repo's `pr-preview` environment.
2. Re-dispatch a recent FE PR preview against this branch.
3. SSH into the deployed backend container on the RPi5 and inspect the rendered `.env` file. Confirm:
   - Every renamed/reclassified item has a non-empty value
   - Database connects successfully
   - Backend starts and serves requests on the expected port

**If the FE-side `vars.*` resolution probe fails** (FE flagged this risk in the board reply): roll back the items that fail by re-adding them as `secrets` in this PR, leaving the rename in place. The naming consistency is preserved; the reclassification scope shrinks.

### Concerns

- **Cross-repo `vars.*` resolution has not been exercised before in this reusable workflow.** Per FE's reply on the coordinator board, `vars.*` should resolve from the caller's environment the same way environment-scoped secrets do (independent of `secrets: inherit`, which only covers secrets). The smoke test above is the verification — don't merge without it.
- **Tight coordination window required.** The GitHub-environment secret/var rename in the FE repo must happen close in time to this PR's merge. If FE renames before this PR merges, the still-deployed workflow references stale `PR_PREVIEW_*` names that no longer exist → PR preview deploys break until the merge. If this PR merges before FE renames, the new bare names don't resolve → same break. Plan a coordinated window.
- **No production impact.** This change only touches the PR preview workflow; production's `deploy_to_do.yml` is unchanged. Worst case if this PR causes preview-env breakage, prod stays healthy.


lmcdonough (approved):
LGTM 👍

lmcdonough (approved):
LGTM 👍

Approving this, but flagging two things I want us to track before/at merge.

**1. `POSTGRES_DB` and `POSTGRES_SCHEMA` have no `|| 'UNUSED'` fallback (lines 1141–1142).**

Every other field in the `.env` heredoc falls back to `'UNUSED'` (or a sensible default like `'4000'` / `'localhost'`) when the var resolves empty. These two don't. If the `pr-preview` environment in either repo hasn't been renamed yet at deploy time, the var lookup returns an empty string, and we'd write `POSTGRES_DB=''` into the `.env`. That produces a malformed `DATABASE_URL` and the migrator fails with a confusing connection error instead of a clear "env var not set" signal.

Small defensive change worth considering: add `|| 'UNUSED'` to both lines so a misconfigured deploy fails loudly with the same shape as every other unset value. Not blocking, but it would save someone a debugging session.

**2. Lockstep coordination window with the FE repo.**

The description already covers this, but I want it captured as a merge checkbox so we don't lose it:

- [ ] FE renames secrets/vars in their `pr-preview` GitHub environment to the new bare names
- [ ] Backend renames secrets/vars in this repo's `pr-preview` GitHub environment to the new bare names
- [ ] Merge this PR
- [ ] Dispatch one FE PR preview as a smoke test, SSH into the deployed container, confirm the rendered `.env` has non-empty values for every renamed item and the backend is serving requests

If FE renames before this PR merges, the still-deployed workflow looks for `PR_PREVIEW_*` names that no longer exist and previews break. If this PR merges before FE renames, the new bare names don't resolve and previews break the same way. Worth confirming with FE on a specific window before clicking merge.

Everything else is mechanically clean and the names line up with production exactly.

lmcdonough:
One more thing while you resolve the conflict with main.

Commit `41db3bd` ("Pass RECALL_AI_* env vars through PR preview deployment") landed on main after this PR opened. It added three new `PR_PREVIEW_*` entries that need the same treatment as everything else in this PR:

**In the `secrets:` input block**, rename:
- `PR_PREVIEW_RECALL_AI_API_KEY` → `RECALL_AI_API_KEY`
- `PR_PREVIEW_RECALL_AI_REGION` → `RECALL_AI_REGION`
- `PR_PREVIEW_RECALL_AI_WEBHOOK_SECRET` → `RECALL_AI_WEBHOOK_SECRET`

**In the `.env` heredoc**:
```yaml
RECALL_AI_API_KEY='${{ secrets.RECALL_AI_API_KEY || 'UNUSED' }}'
RECALL_AI_REGION='${{ vars.RECALL_AI_REGION || 'us-west-2' }}'
RECALL_AI_WEBHOOK_SECRET='${{ secrets.RECALL_AI_WEBHOOK_SECRET || 'UNUSED' }}'
```

I moved `RECALL_AI_REGION` to `vars.*` since it's a non-sensitive subdomain string, same rule as the other classifications in this PR. The API key and webhook secret stay as `secrets.*`.

Also worth adding to the lockstep checklist: both repos' `pr-preview` GitHub environments need the same three new names (`RECALL_AI_API_KEY`, `RECALL_AI_REGION`, `RECALL_AI_WEBHOOK_SECRET`) added during the rename window.

PR #328, opened 2026-05-24
Improve Recall.ai webhook and OAuth setup docs

jhodapp:
## Description
Improves the Recall.ai webhook and OAuth sections in `docs/setup.md` based on issues hit during a local transcription test run. The current docs technically have the right URL form but bury it in a list, and the `OAUTH_SUCCESS_REDIRECT_URI` naming has a typo trap because clap `String` fields silently fall back to the default when the env var name is misspelled.

### Changes
* Highlight that the `/webhooks/recall_ai` path is required: setting the Recall.ai endpoint to just the ngrok host (e.g. `https://abc123.ngrok-free.dev`) makes every delivery fail with **405 Method Not Allowed**, because the Axum router falls back to a `ServeDir` static handler that only accepts GET/HEAD. Wrong-vs-right examples included.
* Add a new `#### Verify the endpoint` subsection with a `curl -i -X POST` one-liner and a response-code legend: **401** means the request reached the handler (path correct, signature absent); **405** means the path is wrong and hit the static fallback; **404** or no response means ngrok or the backend itself is down.
* Update the ngrok command block to show `--url` for reserved domains (ngrok deprecated `--domain`). The free-plan `ngrok http 4000` form is preserved for users on ephemeral URLs.
* Add a "no `GOOGLE_` prefix" reminder next to `OAUTH_SUCCESS_REDIRECT_URI` in all four env blocks where it appears (dev section, dev full snippet, prod section, prod full snippet). A typo'd `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI` is silently ignored because clap `String` fields with `default_value` fall back to the default without warning.

### Testing Strategy
Documentation-only change. Verified manually by following the updated steps: a previously failing webhook delivery (HTTP 405) became HTTP 200 (or 401 for a no-signature `curl` probe) once the endpoint URL was corrected to include `/webhooks/recall_ai`. No code paths changed; no tests required.

### Concerns
None. The doc reorganises and clarifies prose without renaming any env var or changing any default. Readers who already had a working setup are unaffected.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -197,7 +197,6 @@ Key differences from local setup:
 GOOGLE_CLIENT_ID=<client-id>.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=<client-secret>
 GOOGLE_REDIRECT_URI=https://api.myrefactor.com/api/auth/google/callback
-# Reminder: no `GOOGLE_` prefix on this var (a typo'd `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI` is silently ignored).
 OAUTH_SUCCESS_REDIRECT_URI=https://myrefactor.com/settings
 ```
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -60,6 +60,10 @@ ENCRYPTION_KEY=<64-hex-char output from above>
 GOOGLE_CLIENT_ID=<client-id>.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=<client-secret>
 GOOGLE_REDIRECT_URI=http://localhost:4000/api/auth/google/callback
+
+# Note: this var has NO `GOOGLE_` prefix even though it's used by the Google OAuth flow.
+# A typo like `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI=...` is silently ignored (the code
+# falls back to the default `http://localhost:3000/settings`).
 OAUTH_SUCCESS_REDIRECT_URI=http://localhost:3000/settings
 
 # These have working defaults — only set if you need to override:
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -118,6 +150,7 @@ ENCRYPTION_KEY=<output of: openssl rand -hex 32>
 GOOGLE_CLIENT_ID=<client-id>.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=<client-secret>
 GOOGLE_REDIRECT_URI=http://localhost:4000/api/auth/google/callback
+# Reminder: no `GOOGLE_` prefix on this var (a typo'd `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI` is silently ignored).
```
jhodapp:
```suggestion
```

docs/setup.md
```diff
@@ -165,6 +198,7 @@ Key differences from local setup:
 GOOGLE_CLIENT_ID=<client-id>.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=<client-secret>
 GOOGLE_REDIRECT_URI=https://api.myrefactor.com/api/auth/google/callback
+# Reminder: no `GOOGLE_` prefix on this var (a typo'd `GOOGLE_OAUTH_SUCCESS_REDIRECT_URI` is silently ignored).
```
jhodapp:
```suggestion
```

docs/setup.md
```diff
@@ -192,6 +226,7 @@ ENCRYPTION_KEY=<64-hex-char key from secrets manager>
...
```
jhodapp:
```suggestion
```

PR #329, opened 2026-05-24
Fix PG enum cast in atomic claims, clean up webhook error logging

jhodapp:
## Description
The Recall.ai `recording.done` webhook was returning HTTP 500 in local test runs. Root cause: both atomic-claim functions (`try_claim_completed` in `meeting_recording` and `try_claim_for_processing` in `transcription`) used `Entity::update_many().col_expr(Column::Status, Expr::value(enum_value))`. That `col_expr` path loses the column's PG enum type metadata at bind time, so sqlx sent the value as `text` and PostgreSQL rejected the assignment with error `42804`:

```
column "status" is of type meeting_recording_status but expression is of type text
```

The same bug existed in the transcription claim path but was latent; the failure happened earlier in the webhook flow.

Bundled with the bug fix: a small cleanup of the deeply-nested error log the webhook handler produces on failure. `domain::Error` and `entity_api::Error` `Display` impls were delegating straight to `Debug`, so `{}` and `{:?}` produced identical noise.

### Changes
* Rewrite `try_claim_completed` in `entity_api/src/meeting_recording.rs` to use `db.transaction(...)` with a `SELECT ... FOR UPDATE` row lock followed by `ActiveModel.update()`. The `Set(enum_variant)` path uses SeaORM's column metadata to bind the parameter as the proper PG enum type, so no `CAST` is needed in the SQL. This restores the codebase invariant that every PG enum write goes through `ActiveModel + Set()`.
* Apply the same rewrite to `try_claim_for_processing` in `entity_api/src/transcription.rs`.
* Fix `Display` on `domain::Error` in `domain/src/error.rs` to print ``{error_kind:?}: {source}`` and recurse through `source` via `Display`. Was: ``"Domain Error: {self:?}"``.
* Fix `Display` on `entity_api::Error` in `entity_api/src/error.rs` to do the same. Was: ``"Entity API Error: {self:?}"``.
* Switch `web/src/controller/webhook_controller.rs` from ``error!("...: {:?}", e)`` to ``error!("...: {e}")``.
* Update both functions' mock tests to match the SELECT-then-UPDATE shape. Each function gains an explicit "not found" test case (the previous implementation conflated "row missing" with "row already terminal"; both returned 0 rows affected. The new implementation distinguishes the two).

#### Automated
* `cargo test --package entity_api --features mock --lib`: 174 passed, 0 failed.
* `cargo test --package domain`: 1 passed (others ignored as before).
* `cargo check --workspace`: clean.
* `cargo clippy` on the edited files: no new warnings.
* `cargo fmt --check`: clean.

#### End-to-end (already performed against a real PostgreSQL)
* Replayed a failed `recording.done` delivery from the Recall.ai dashboard against the local backend after applying the enum-cast fix.
* Confirmed HTTP 200, recording status flipped from `processing` to `completed`, and the transcription job kicked off (also exercising `try_claim_for_processing`).

#### Before / after on the webhook error log

Before (single line, fully unwrapped here for readability):
```
[ERROR] Recall.ai webhook error: Error { source: Some(Error { source: Some(Exec(SqlxError(Database(PgDatabaseError { severity: Error, code: "42804", message: "column \"status\" is of type meeting_recording_status but expression is of type text", detail: None, hint: Some("You will need to rewrite or cast the expression."), position: Some(Original(64)), where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("parse_target.c"), line: Some(593), routine: Some("transformAssignedExpr") })))), error_kind: SystemError }), error_kind: Internal(Entity(ServiceUnavailable)) }
```

After:
```
[ERROR] Recall.ai webhook error: Internal(Entity(ServiceUnavailable)): SystemError: Execution Error: error returned from database: column "status" is of type meeting_recording_status but expression is of type text
```

### Concerns & Notes
* The original `update_many` form was one round-trip. The new form is two queries inside a transaction (`SELECT FOR UPDATE`, then `UPDATE ... RETURNING`). Concurrent webhook deliveries for the same row now serialize behind the row lock instead of racing at the UPDATE level. For a function called at most a handful of times per coaching session, this is the right trade. Atomicity stays correct, and the failure mode (enum cast bug) is gone.
* The `Display` change is technically a behavior change for any code that called `format!("{e}")` on a `domain::Error` or `entity_api::Error`. Grep shows no callers depend on the old `"Domain Error: ..."` or `"Entity API Error: ..."` text. Anything using `{:?}` is unaffected.

### DIFF
--- a/domain/src/error.rs
+++ b/domain/src/error.rs
@@ -72,7 +72,11 @@ pub enum ExternalErrorKind {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Domain Error: {self:?}")
+        write!(f, "{:?}", self.error_kind)?;
+        if let Some(ref src) = self.source {
+            write!(f, ": {src}")?;
+        }
+        Ok(())
     }
 }
 
--- a/entity_api/src/error.rs
+++ b/entity_api/src/error.rs
@@ -45,7 +45,11 @@ pub enum EntityApiErrorKind {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Entity API Error: {self:?}")
+        write!(f, "{:?}", self.error_kind)?;
+        if let Some(ref src) = self.source {
+            write!(f, ": {src}")?;
+        }
+        Ok(())
     }
 }
 
--- a/entity_api/src/meeting_recording.rs
+++ b/entity_api/src/meeting_recording.rs
@@ -5,8 +5,15 @@ use log::debug;
 use sea_orm::{
     entity::prelude::*,
     ActiveValue::{Set, Unchanged},
-    DatabaseConnection, Order, QueryOrder, TryIntoModel,
+    DatabaseConnection, IntoActiveModel, Order, QueryOrder, QuerySelect, TransactionError,
+    TransactionTrait, TryIntoModel,
 };
+
+const TERMINAL_RECORDING_STATUSES: &[MeetingRecordingStatus] = &[
+    MeetingRecordingStatus::Completed,
+    MeetingRecordingStatus::Failed,
+    MeetingRecordingStatus::Cancelled,
+];
 
 /// Creates a new meeting recording record
 pub async fn create(db: &DatabaseConnection, model: Model) -> Result<Model, Error> {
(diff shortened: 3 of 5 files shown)


domain/src/error.rs
```diff
@@ -72,7 +72,11 @@ pub enum ExternalErrorKind {
...
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Domain Error: {self:?}")
+        write!(f, "{:?}", self.error_kind)?;
+        if let Some(ref src) = self.source {
+            write!(f, ": {src}")?;
+        }
+        Ok(())
     }
 }
```
jhodapp, line 81: This and the other one should result in much better looking error messages in the logs, something that Claude caught while working on this fix since Recall's error messages were very deeply nested and not very helpful to read. This is more idiomatic the Display trait's `fmt` implement anyway.

I'm going to merge this as-is, but if you disagree with any of the changes here let's do a follow-up PR.

Issue #333, opened 2026-05-25
Attach .ics calendar invite to coaching-session emails (scheduled, rescheduled, recurring, cancel)


jhodapp:
## Summary
Attach an RFC 5545 `.ics` file to the transactional emails sent for coaching-session lifecycle events so coach and coachee can add, update, or remove the meeting in Apple Calendar, Google Calendar, and Outlook with one click. The lifecycle spans seven scenarios across two cardinalities (single session, series) and the actions schedule, update/reschedule, cancel, and per-occurrence cancel.

> **Implementation plan:** [`docs/implementation-plans/ics-calendar-invites-coaching-sessions.md`](https://github.com/refactor-group/refactor-platform-rs/blob/main/docs/implementation-plans/ics-calendar-invites-coaching-sessions.md)

## Scope phasing (foundation-first)
- **Core foundation:** one **reusable `.ics` builder** that can emit *every* scenario shape, built and unit-tested before anything consumes it. The scenarios are populations of one input struct varying along three orthogonal axes — METHOD (`REQUEST`/`CANCEL`) × cardinality (single `VEVENT` vs `VEVENT`+`RRULE`) × `SEQUENCE` — plus, for the per-occurrence case, an `EXDATE`/`RECURRENCE-ID` exception.
- **Attach to existing create emails first:** the new-single and new-series flows already send emails; the first consuming phase just adds the `.ics` attachment.
- **Net-new emails follow as sequential next steps:** update/reschedule, cancel, and per-occurrence cancel currently send no email. Each lands in its own later phase once the foundation and create-flow attachment are proven.
- Each phase is an independently reviewable PR that **merges to `main` one at a time**; phase N+1 branches from `main` after phase N lands.

### PR #356 — `coaching_session_series` (hard dependency)
This work must be built **on top of [#356](https://github.com/refactor-group/refactor-platform-rs/pull/356)** (branch from it or land after it merges). #356 makes a recurring series a first-class entity and changes the surfaces this issue hooks into:
- New `domain::coaching_session_series` with typed `SeriesRule { start_at, recurrence, duration_minutes }` and `create_with_sessions` / `reschedule` / `delete_with_future_sessions`.
- New `coaching_session_series` entity (stable `id`, `rule` JSONB) + `coaching_sessions.coaching_session_series_id` FK.
[...]

PR #334, opened 2026-05-25
Add duration_minutes to coaching_sessions and per-coach default

### LINKED ISSUE #332: Add duration_minutes to coaching_sessions
## Summary
Add a `duration_minutes` field to `coaching_sessions` so the platform has an explicit session length, plus a per-coach `default_coaching_session_duration_minutes` preference on `users` so each coach can configure their own default. Today `coaching_sessions` carries `date` (start) but no duration, forcing every consumer that needs an end time to assume one. This is a prerequisite for attaching `.ics` calendar invites to session emails (#333).

The duration is also surfaced in the existing session-scheduled and recurring-sessions-scheduled emails so recipients see how long their session is.

## Scope

**Migrations** ([migration/](migration/))
- Add `duration_minutes INTEGER NOT NULL DEFAULT 60` to `refactor_platform.coaching_sessions`. The default backfills existing rows with 60 — the historical norm for one-on-one coaching.
- Add `default_coaching_session_duration_minutes INTEGER NOT NULL DEFAULT 60` to `refactor_platform.users`. The column is universal across all users; only the Coach role meaningfully uses it, but no role-gating is enforced at the schema level (mirrors how `timezone` is set up today).

**Entity** ([entity/src/coaching_sessions.rs](entity/src/coaching_sessions.rs), [entity/src/users.rs](entity/src/users.rs))
- Regenerate both to include the new fields.

**Domain layer** ([domain/src/coaching_session.rs](domain/src/coaching_session.rs), [domain/src/emails.rs](domain/src/emails.rs), [domain/src/user.rs](domain/src/user.rs))
- Validate `duration_minutes` on session `create` and `update`: must be in `1..=480` (1 min to 8 hours). Reject out-of-range input.
- Validate `default_coaching_session_duration_minutes` on user update with the same `1..=480` bounds.
- **Defaulting cascade for session create paths:** when `duration_minutes` is omitted from the request, resolve in this order — coach's `default_coaching_session_duration_minutes` → hardcoded 60 fallback. Applies to both `coaching_session::create` and `bulk_create_recurring`.
- `bulk_create_recurring`: every materialized session inherits the resolved duration.
- New helper `format_duration(minutes: i32) -> String` alongside `format_session_date_time` ([emails.rs:254](domain/src/emails.rs#L254)). Produces human strings: `"1 hour"`, `"45 minutes"`, `"1 hour 30 minutes"`.
- `send_session_email_to_recipient` ([emails.rs:348](domain/src/emails.rs#L348)) and `send_recurring_series_email_to_recipient` ([emails.rs:537](domain/src/emails.rs#L537)) pass a new template variable `session_duration` to Resend.

**Web layer**
- `POST /coaching_sessions`: accept optional `duration_minutes` in the body; if omitted, defaulting cascade applies.
- `POST /coaching_sessions/recurring`: `CreateRecurringParams` accepts optional `duration_minutes`, applied to every session in the series (defaulting cascade applies when omitted).
- `PUT /coaching_sessions/{id}`: `duration_minutes` is patchable.
- `GET` responses for coaching sessions include `duration_minutes`.
- `GET /users/{id}` (and equivalent self/me endpoints) include `default_coaching_session_duration_minutes`.
- User update endpoint accepts `default_coaching_session_duration_minutes` as a patchable field.
- `utoipa` schemas updated for both new fields.

**Resend templates (manual)**
Update the **session-scheduled** and **recurring-sessions-scheduled** templates in the Resend dashboard to render the new `session_duration` variable in the email body. Welcome, password-reset, and action-assigned templates are unaffected.

**Tests**
- Migration backfills existing coaching-sessions rows to 60 and existing users rows to 60.
- Out-of-range duration is rejected on session create, session update, and user update.
- `POST /coaching_sessions` without `duration_minutes` resolves to the coach's `default_coaching_session_duration_minutes`.
- A coach with a non-default `default_coaching_session_duration_minutes` (e.g. 45) gets that value applied to a session created without an explicit duration.
- Every recurring-series session carries the resolved duration.
- `format_duration` unit tests for 1, 45, 60, 90, 480.
- Existing `mockito` wire-contract tests at [emails.rs:1052](domain/src/emails.rs#L1052) and [emails.rs:1492](domain/src/emails.rs#L1492) updated to expect `session_duration` in the payload.

## Frontend coordination
The frontend will file a paired issue (coordinated via the shared collab blackboard) for the UI changes:
- Duration input in the schedule/reschedule dialog.
- New user-settings UI to configure `default_coaching_session_duration_minutes`.

The BE work in this issue can ship independently — sessions created via the existing dialog (no `duration_minutes` in the payload) will simply use the coach's default.

## Out of scope
- Variable duration within a single recurring series.
- Per-relationship duration overrides (one coach, multiple coachees with different default lengths).
- Frontend UI changes ([separate FE issue](https://github.com/refactor-group/refactor-platform-fe/issues/402)).

## Acceptance criteria
- [ ] Migration adds `duration_minutes INTEGER NOT NULL DEFAULT 60` to `coaching_sessions`; existing rows backfilled to 60.
- [ ] Migration adds `default_coaching_session_duration_minutes INTEGER NOT NULL DEFAULT 60` to `users`; existing rows backfilled to 60.
- [ ] `POST /coaching_sessions` without `duration_minutes` resolves to the coach's `default_coaching_session_duration_minutes` (not a hardcoded constant).
- [ ] A coach whose `default_coaching_session_duration_minutes` is 45 creates a session without specifying duration and the resulting session has `duration_minutes = 45`.
- [ ] `duration_minutes` outside `1..=480` returns 4xx on session create and update.
- [ ] `default_coaching_session_duration_minutes` outside `1..=480` returns 4xx on user update.
- [ ] `PUT /coaching_sessions/{id}` can patch only `duration_minutes` without touching other fields.
- [ ] Every session from `POST /coaching_sessions/recurring` carries the resolved duration.
- [ ] `GET` responses for sessions include `duration_minutes`; user GET responses include `default_coaching_session_duration_minutes`.
- [ ] Session-scheduled and recurring-sessions-scheduled emails include `session_duration` in their Resend payload, formatted as a human string.
- [ ] Existing wire-contract tests are updated to expect the new variable; full test suite passes.

jhodapp:
## Description
Adds an explicit duration to coaching sessions and a per-coach default preference, plus a defaulting cascade so omitting `duration_minutes` from a session-create payload resolves to the coach's stored preference. Surfaces the duration in existing session-scheduled and recurring-sessions-scheduled emails. Implementation plan committed at [docs/implementation-plans/coaching-session-duration.md](docs/implementation-plans/coaching-session-duration.md).

### Changes
* Two migrations: `coaching_sessions.duration_minutes SMALLINT NOT NULL DEFAULT 60` and `users.default_coaching_session_duration_minutes SMALLINT NOT NULL DEFAULT 60` — both backfill existing rows.
* New `entity::duration::Duration(u16)` newtype with smart constructor (`Duration::new`, `TryFrom<u16>`), `Display` impl covering all eight singular/plural cases, `Default` impl returning 60 minutes, `from_minutes_unchecked` escape hatch for trusted DB reads, and `OutOfRange` error type. Range invariant: `1..=480`.
* Re-exported through `entity_api` and `domain` at the `lib.rs` level (matches the pattern for `Provider`, `Id`, etc.). New "Type Re-Export Boundary" subsection added to `.claude/coding-standards.md` documenting this convention.
* Defaulting cascade in `entity_api::coaching_session::resolve_duration`: `Some(d)` uses the value directly, `None` loads the coach and returns their default.
* New `CreateParams` struct decouples the `POST /coaching_sessions` request body from the entity Model so `duration_minutes` can be optional on the wire while remaining non-null in storage. Wire-shape transparent for existing FE payloads (the entity Model already had `#[serde(skip_deserializing)]` on the fields `CreateParams` omits).
* Controllers convert `Option<u16>` → `Option<Duration>` at the boundary; out-of-range values propagate as **422 `validation_error`** consistently across POST, PUT, recurring, and user-update endpoints. New `From<OutOfRange> for domain::Error` impl routes through `DomainErrorKind::Validation`.
* Email payload variable `{{session_duration}}` added to `send_session_email_to_recipient` and `send_recurring_series_email_to_recipient`, formatted via `Duration`'s `Display` impl.

### Testing Strategy
* `cargo test --workspace` — 18 new unit tests for the `Duration` newtype; existing wire-contract tests in `domain/src/emails.rs` updated to expect the new `session_duration` variable; 11 SQL-emission tests in `entity_api` updated to expect the new columns in the SELECT output.
* `cargo clippy --workspace --all-targets -- -D warnings` — clean.
* Manual smoke (post-deploy + Resend template update):
  1. `POST /coaching_sessions` with `duration_minutes: 45` → confirm the resulting session has 45-minute duration and the email contains "for 45 minutes".
  2. `POST /coaching_sessions` **omitting** `duration_minutes` when the coach's `default_coaching_session_duration_minutes` is set to a non-default value → confirm the session uses the coach's default.
  3. `PUT /coaching_sessions/{id}` with `duration_minutes: 481` → confirm 422 `validation_error` with message including the offending value.
  4. `PUT /users/{id}` with `default_coaching_session_duration_minutes: 481` → confirm 422 `validation_error`.
  5. `POST /coaching_sessions/recurring` with `duration_minutes: 30` → confirm every materialized session has 30 minutes.

### Concerns
* **Resend dashboard templates need a one-time update after this merges + deploys.** Two templates (`session-scheduled` and `recurring-sessions-scheduled`) need `{{session_duration}}` rendered in their body — until then, the variable is sent but not displayed. No 5xx risk; emails still send.
* **Frontend coordination is on the collab board.** Contract `CoachingSessionDurationFeature v2` pins the wire format the FE should type against; the FE issue is unblocked. No flag-gating needed — the BE defaulting cascade makes omitting `duration_minutes` always safe, so the FE can roll out the dialog input on its own timeline.
* **Unblocks #333** (`.ics` calendar invites), which needed `duration_minutes` to compute `DTEND`.


domain/src/error.rs
```diff
@@ -88,6 +88,19 @@ impl StdError for Error {
...
+impl From<entity::duration::OutOfRange> for Error {
+    fn from(err: entity::duration::OutOfRange) -> Self {
+        let message = err.to_string();
+        Error {
+            source: Some(Box::new(err)),
+            error_kind: DomainErrorKind::Validation(message),
+        }
+    }
+}
+
```
jhodapp: @greptileai A note to remember for your reviews, this is wrong. This belongs as an addition to the `from()` impl for `EntityApiError` on line 106. I want you to catch future violations like this.

PR #345, opened 2026-06-05
Meeting provider refactor introducing cost observability

calebbourg:
## Description
Introduces cost observability for Recall.ai usage by laying the full data and domain foundation: DB schema, entity definitions, CRUD modules, and live cost recording wired into the existing webhook handlers.

The work comes in three layers:
1. **Schema** — renames the existing `provider` enum to `meeting_provider` (PostgreSQL tracks column OIDs so no `ALTER TABLE` needed), adds a separate `pipeline_provider` enum (`recall_ai`, `llm_gateway`), and introduces a `cost_pricing_config` rate-history table and a `platform_cost_metrics` per-session cost table.
2. **Entity/domain refactor** — separates `meeting_provider` (DB-backed, used by `coaching_sessions` and `oauth_connections`) from `pipeline_provider` (AI pipeline layer), keeps a Rust-only `provider::Provider` wrapper for generic domain logic, adds entity and CRUD modules for all new tables, and removes the dead `AssemblyAi` variant from `meeting-auth` (Recall.ai handles AssemblyAI internally).
3. **Cost recording** — wires best-effort cost capture into the webhook handlers: `recording.done` records bot minutes, `transcript.done` records transcription hours. Both look up the current rate from `cost_pricing_config` and write a `platform_cost_metrics` row. If no rate is configured they warn and return `Ok` so the webhook always succeeds.

### Changes
* Two new migrations: rename `provider` → `meeting_provider`; add `pipeline_provider`, `cost_metric`, `cost_unit`, `cost_pricing_config`, and `platform_cost_metrics` tables
* New entities: `meeting_provider`, `pipeline_provider`, `cost_metric`, `cost_unit`, `cost_pricing_config`, `platform_cost_metrics`
* New `entity_api` modules: `cost_pricing_config` (CRUD), `platform_cost_metrics` (CRUD)
* New `domain/src/cost.rs`: `record_bot_minutes` and `record_transcription_hours` best-effort functions
* `recording.done` and `transcript.done` webhook handlers wired to spawn cost recording after a successful claim
* All `provider::Provider` call sites updated to `meeting_provider::Provider` where a DB-backed meeting provider is expected
* Removed dead `AssemblyAi` variant from `meeting-auth/src/api_key/auth.rs`
* Fixed pre-existing mock test bugs in both webhook handlers where an empty query queue caused `MockDatabase` to error instead of returning `Ok(None)`

### Testing Strategy
* `cargo test` — existing unit tests plus new tests in `domain/src/cost.rs` covering happy paths and both no-op cases (no rate configured, recording not found); mock tests in `recording_done.rs` and `transcript_done.rs` verify the claim-skip paths
* Manually configure a `cost_pricing_config` row for `(RecallAi, BotMinutes)` and `(RecallAi, TranscriptionHours)`, trigger a recording + transcript webhook, and confirm rows appear in `platform_cost_metrics`
* Trigger the webhooks with no matching pricing row to confirm the handlers still return `200 OK` and log a warning

### Concerns
* Cost recording is best-effort and fire-and-forget (spawned task). A failed insert is only logged — there is no retry or alerting mechanism yet.
* `cost_pricing_config` has no UI or admin endpoint; rows must be seeded directly in the database for now.


domain/src/cost.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    platform_cost_metrics::create(
+        db,
+        CostMetricsModel {
+            id: Id::new_v4(),
+            provider: Provider::RecallAi,
+            metric: Metric::TranscriptionHours,
+            coaching_session_id: Some(recording.coaching_session_id),
+            source_record_id: transcription_id,
+            cost_low: quantity * rate.cost_per_unit_low,
```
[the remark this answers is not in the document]
jhodapp: @calebbourg just surfacing to make sure you see Greptile's review comments.

domain/src/webhook/transcript_done.rs
```diff
@@ -67,6 +68,16 @@ pub async fn handle(
...
 
+        if let Err(e) =
+            crate::cost::record_transcription_hours(&db, transcription_id, meeting_recording_id)
+                .await
+        {
+            warn!(
+                "cost: transcription hours failed for transcription {}: {:?}",
+                transcription_id, e
+            );
+        }
```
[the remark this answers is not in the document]
jhodapp: @calebbourg just surfacing to make sure you see Greptile's review comments.

entity/src/provider.rs
```diff
@@ -1,58 +1,35 @@
...
-)]
-#[serde(rename_all = "lowercase")]
-#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "provider")]
-#[schema(as = domain::provider::Provider)]
+/// Rust-only wrapper enum for any provider in the system.
+/// Not backed by a DB column — use `meeting_provider::Provider` or
+/// `pipeline_provider::Provider` for DB fields.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+#[serde(tag = "type", content = "provider", rename_all = "snake_case")]
 pub enum Provider {
```
jhodapp: `provider::Provider` wrapper is now dead code. The whole type, its two `From` impls, and the `Display` impl are reachable only through `Provider`, which nothing constructs or consumes. It's still re-exported through `entity_api` and `domain` and derives `ToSchema` without being registered in OpenAPI. I recommend deleting it and we re-introduce it when a generic path needs it, although I'm curious for your thoughts on this.

entity/src/cost_pricing_config.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct Model {
+    #[serde(skip_deserializing)]
+    #[sea_orm(primary_key)]
+    pub id: Id,
+    pub provider: pipeline_provider::Provider,
+    pub metric: cost_metric::Metric,
+    pub unit: cost_unit::Unit,
+    pub cost_per_unit_low: f64,
+    pub cost_per_unit_high: f64,
+    pub cost_per_unit_avg: f64,
```
jhodapp: Monetary values stored as `f64` (`DOUBLE PRECISION`). These get summed for reporting, where floating-point drift accumulates — prefer Postgres `NUMERIC`. Suggest different scales for the two roles:

- Per-unit rates (`cost_pricing_config.cost_per_unit_*`): `NUMERIC(20, 10)` — per-token pricing can be ~`$0.0000025`, so it needs high fractional precision or the rate rounds to zero before multiplication.
- Computed costs (`platform_cost_metrics.cost_*`): `NUMERIC(14, 6)` — micro-dollar precision (standard for usage-based billing), with room to ~$100M per row.

entity/src/platform_cost_metrics.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub id: Id,
+    pub provider: pipeline_provider::Provider,
+    pub metric: cost_metric::Metric,
+    pub coaching_session_id: Option<Id>,
+    /// Logical FK to the source record (meeting_recordings.id or transcriptions.id).
+    /// Not a DB-level FK because it can point to different tables.
+    pub source_record_id: Id,
+    pub cost_low: f64,
+    pub cost_high: f64,
+    pub cost_avg: f64,
```
jhodapp: Same `f64` concern — prefer `NUMERIC`. These are computed dollar totals, so `NUMERIC(14, 6)` (micro-dollar precision) is good enough.

migration/src/m20260529_000001_add_cost_tables.rs
```diff
@@ -0,0 +1,167 @@
...
+            .execute_unprepared(
+                r#"
+                CREATE TABLE IF NOT EXISTS refactor_platform.cost_pricing_config (
+                    id                  UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+                    provider            refactor_platform.pipeline_provider NOT NULL,
+                    metric              refactor_platform.cost_metric NOT NULL,
+                    unit                refactor_platform.cost_unit NOT NULL,
+                    cost_per_unit_low   DOUBLE PRECISION NOT NULL,
+                    cost_per_unit_high  DOUBLE PRECISION NOT NULL,
+                    cost_per_unit_avg   DOUBLE PRECISION NOT NULL,
```
jhodapp: `DOUBLE PRECISION` drifts when summed; prefer `NUMERIC`. Rates (lines 59-61): `NUMERIC(20, 10)`. Computed costs (lines 88-90): `NUMERIC(14, 6)`. Cheap to change now but once rows exist, `ALTER COLUMN TYPE` needs a rewrite lock.

domain/src/cost.rs
```diff
@@ -0,0 +1,270 @@
...
+        None => {
+            warn!(
+                "cost: recording {} not found — skipping bot minutes cost",
+                recording_id
+            );
+            return Ok(());
+        }
+    };
+
+    let quantity = recording.duration_seconds.unwrap_or(0) as f64 / 60.0;
```
jhodapp: `recording.duration_seconds.unwrap_or(0)` persists a zero-cost row when duration is `None`/`0`, polluting aggregates and contradicting this module's own "warn + no-op" style (missing-rate, missing-recording branches). I suggest a warn-and-skip approach instead of inserting. What do you think?

domain/src/cost.rs
```diff
@@ -0,0 +1,270 @@
...
+        None => {
+            warn!(
+                "cost: recording {} not found — skipping transcription hours cost",
+                meeting_recording_id
+            );
+            return Ok(());
+        }
+    };
+
+    let quantity = recording.duration_seconds.unwrap_or(0) as f64 / 3600.0;
```
jhodapp: Same as the bot-minutes path: `unwrap_or(0)` writes a `0.0` transcription-hours row when duration is missing. Warn-and-skip rather than insert a meaningless zero row.

domain/src/webhook/recording_done.rs
```diff
@@ -70,6 +70,17 @@ pub async fn handle(
...
+    let recording_id = recording.id;
+    let db_cost = Arc::clone(&db);
+    tokio::spawn(async move {
+        if let Err(e) = crate::cost::record_bot_minutes(&db_cost, recording_id).await {
+            warn!(
+                "cost: bot minutes failed for recording {}: {:?}",
+                recording_id, e
+            );
+        }
+    });
```
jhodapp: `recording.done` now spawns two detached tasks (bot-minutes recording + transcription start), each grabbing its own pooled connection. Given the known DB pool-exhaustion issue (churn → `ConnectionAcquire(Timeout)`), consider folding this single INSERT into the existing transcription spawn sequentially. Also note the shape differs from `transcript_done`, which awaits cost inline.

domain/src/webhook/transcript_done.rs
```diff
@@ -67,6 +68,16 @@ pub async fn handle(
             .await;
         }
 
+        if let Err(e) =
+            crate::cost::record_transcription_hours(&db, transcription_id, meeting_recording_id)
```
jhodapp, line 74: `record_transcription_hours` runs after the `if let Err(e) = result` block, not inside the success branch, so a failed transcription still records cost. If Recall.ai bills for the attempt regardless this is fine, but it's currently implicit. Add a one-line comment stating the intent.

migration/src/m20260529_000001_add_cost_tables.rs
```diff
@@ -0,0 +1,167 @@
...
+                "ALTER TABLE refactor_platform.cost_pricing_config OWNER TO refactor",
+            )
+            .await?;
+
+        // platform_cost_metrics: one row per session per provider+metric, written by webhook handlers
+        manager
+            .get_connection()
+            .execute_unprepared(
+                r#"
+                CREATE TABLE IF NOT EXISTS refactor_platform.platform_cost_metrics (
```
jhodapp, line 100: No unique constraint on `platform_cost_metrics`. Idempotency rests entirely on the upstream claim gates. A `UNIQUE (source_record_id, metric)` index would be cheap defense-in-depth if cost recording is ever invoked from a second path (backfill/admin). Contrast `cost_pricing_config`, which has `UNIQUE (provider, metric, effective_from)` at line 63.

migration/src/m20260529_000001_add_cost_tables.rs
```diff
@@ -0,0 +1,167 @@
...
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // pipeline_provider enum: services that form the record → transcribe → analyze chain
+        manager
+            .get_connection()
+            .execute_unprepared(
+                "CREATE TYPE refactor_platform.pipeline_provider AS ENUM ('recall_ai', 'llm_gateway')",
```
jhodapp: `CREATE TYPE` without `IF NOT EXISTS`. Minor robustness note for partial-failure re-runs. Consistent with existing migrations, so non-blocking and just a suggestion for a new pattern moving forward.

migration/src/m20260529_000001_add_cost_tables.rs
```diff
@@ -0,0 +1,167 @@
...
+        manager
+            .get_connection()
+            .execute_unprepared("ALTER TYPE refactor_platform.pipeline_provider OWNER TO refactor")
+            .await?;
+
+        // cost_metric enum: what is being measured for billing purposes
+        manager
+            .get_connection()
+            .execute_unprepared(
+                "CREATE TYPE refactor_platform.cost_metric AS ENUM \
```
jhodapp: `CREATE TYPE` without `IF NOT EXISTS`. Minor robustness note for partial-failure re-runs. Consistent with existing migrations, so non-blocking.

migration/src/m20260529_000001_add_cost_tables.rs
```diff
@@ -0,0 +1,167 @@
...
+        manager
+            .get_connection()
+            .execute_unprepared("ALTER TYPE refactor_platform.cost_metric OWNER TO refactor")
+            .await?;
+
+        // cost_unit enum: the unit of measure for each cost_metric
+        manager
+            .get_connection()
+            .execute_unprepared(
+                "CREATE TYPE refactor_platform.cost_unit AS ENUM ('minutes', 'hours', 'tokens')",
```
jhodapp: `CREATE TYPE` without `IF NOT EXISTS`. Minor robustness note for partial-failure re-runs. Consistent with existing migrations, so non-blocking.

domain/src/cost.rs
```diff
@@ -0,0 +1,270 @@
...
+use entity_api::{cost_pricing_config, meeting_recording as recording_api, platform_cost_metrics};
+use log::warn;
+use sea_orm::DatabaseConnection;
+
+/// Records the Recall.ai bot-minutes cost for a completed recording.
+///
+/// Fetches the recording fresh from the DB so `duration_seconds` reflects the
+/// value written by `try_claim_completed`. No-ops with a warning if no pricing
+/// row is configured.
+pub async fn record_bot_minutes(db: &DatabaseConnection, recording_id: Id) -> Result<(), Error> {
```
jhodapp, line 14: `record_bot_minutes` and `record_transcription_hours` are nearly identical: same rate lookup, same recording fetch, same `quantity * rate.cost_per_unit_*` triple, same insert. The only differences are the metric, the source-record id, and the `/60` vs `/3600` divisor. Recommend extracting a single private helper both delegate to, and replacing the `match Some/None { ... return Ok(()) }` blocks with `let ... else` to match the combinator style used elsewhere (e.g. `find_current_rate`). Sketch:

```rust
async fn record(
    db: &DatabaseConnection,
    metric: Metric,
    source_record_id: Id,
    recording: &CostRecordingModel,
) -> Result<(), Error> {
    let Some(rate) =
        cost_pricing_config::find_current_rate(db, Provider::RecallAi, metric).await?
    else {
        warn!("cost: no pricing config for (RecallAi, {metric}) — skipping {source_record_id}");
        return Ok(());
    };

    let Some(quantity) = rate.unit.quantity_from_seconds(recording.duration_seconds) else {
        return Ok(()); // non-duration metric, nothing to record from a recording
    };

    let cost = rate.cost_for(quantity);
    platform_cost_metrics::create(db, CostMetricsModel {
        id: Id::new_v4(),
        provider: Provider::RecallAi,
        metric,
        coaching_session_id: Some(recording.coaching_session_id),
        source_record_id,
        cost_low: cost.low,
        cost_high: cost.high,
        cost_avg: cost.avg,
        created_at: chrono::Utc::now().fixed_offset(),
    })
    .await?;
    Ok(())
}
```

The two public functions then become thin wrappers that fetch the recording and call `record(...)`.

entity/src/cost_pricing_config.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[schema(value_type = String, format = DateTime)]
+    pub effective_from: DateTimeWithTimeZone,
```
jhodapp, line 31: The `quantity * rate.cost_per_unit_{low,high,avg}` triple is written out twice in `cost.rs` (lines 51-53 and 109-111). Move it onto the rate model as a small value type so callers never re-derive it:

```rust
pub struct CostRange { pub low: f64, pub high: f64, pub avg: f64 }

impl Model {
    pub fn cost_for(&self, quantity: f64) -> CostRange {
        CostRange {
            low:  quantity * self.cost_per_unit_low,
            high: quantity * self.cost_per_unit_high,
            avg:  quantity * self.cost_per_unit_avg,
        }
    }
}
```

This also makes the f64→NUMERIC change from the other review comments a one-place edit later.

entity/src/cost_unit.rs
```diff
@@ -0,0 +1,31 @@
...
+    Deserialize,
+    Serialize,
+    DeriveActiveEnum,
+    Default,
+    ToSchema,
+)]
+#[serde(rename_all = "snake_case")]
+#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "cost_unit")]
+#[schema(as = domain::cost_unit::Unit)]
+pub enum Unit {
```
jhodapp, line 22: The `/60` and `/3600` divisors in `cost.rs` (lines 41, 99) are the seconds-per-unit relationship that *this enum already names*. Give `Unit` the behavior, mirroring how `meeting_provider::Provider` carries `MeetingProperties`:

```rust
impl Unit {
    /// Billable quantity for a duration, or `None` for non-duration units (e.g. tokens).
    pub fn quantity_from_seconds(&self, seconds: Option<i64>) -> Option<f64> {
        let seconds = seconds.filter(|s| *s > 0)? as f64;
        match self {
            Unit::Minutes => Some(seconds / 60.0),
            Unit::Hours   => Some(seconds / 3600.0),
            Unit::Tokens  => None,
        }
    }
}
```

Bonus: returning `None` for missing/zero duration is what folds in the "don't write a zero-cost row" fix from the other review — the helper in #1 just `return Ok(())`s. The divisor is now derived from `rate.unit` rather than hard-coded per function.

jhodapp (changes requested):
Thanks Caleb for the excellent work, as always. Let me know if you have any questions about my PR review feedback.

entity_api/src/cost_pricing_config.rs
```diff
@@ -0,0 +1,108 @@
...
+pub async fn create(db: &DatabaseConnection, model: Model) -> Result<Model, Error> {
+    let active_model = ActiveModel {
+        id: Set(Id::new_v4()),
+        provider: Set(model.provider),
+        metric: Set(model.metric),
+        unit: Set(model.unit),
+        cost_per_unit_low: Set(model.cost_per_unit_low),
+        cost_per_unit_high: Set(model.cost_per_unit_high),
+        cost_per_unit_avg: Set(model.cost_per_unit_avg),
+        effective_from: Set(model.effective_from),
```
jhodapp: `DeriveEntityModel` already generates `IntoActiveModel`, so the field-by-field `Set(...)` spread is boilerplate that drifts whenever a column is added. Prefer:

```rust
pub async fn create(db: &DatabaseConnection, model: Model) -> Result<Model, Error> {
    let mut active = model.into_active_model();
    active.id = Set(Id::new_v4()); // override caller-supplied id
    Ok(active.insert(db).await?)
}
```
(Requires `use sea_orm::IntoActiveModel;`)

entity_api/src/platform_cost_metrics.rs
```diff
@@ -0,0 +1,100 @@
...
+    let active_model = ActiveModel {
+        id: Set(Id::new_v4()),
+        provider: Set(model.provider),
+        metric: Set(model.metric),
+        coaching_session_id: Set(model.coaching_session_id),
+        source_record_id: Set(model.source_record_id),
+        cost_low: Set(model.cost_low),
+        cost_high: Set(model.cost_high),
+        cost_avg: Set(model.cost_avg),
+        created_at: Set(chrono::Utc::now().fixed_offset()),
```
jhodapp: Same as `cost_pricing_config::create`: replace the manual spread with `into_active_model()`, then override only the generated fields (`id`, `created_at`). Removes 8 lines of mechanical `Set(model.x)`.

jhodapp (changes requested):
A couple more I forgot in my last review.

domain/src/webhook/transcript_done.rs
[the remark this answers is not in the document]
calebbourg: d88fcc74baeef7239af4a06a67365b8fa5cc9020

domain/src/cost.rs
calebbourg: d88fcc74baeef7239af4a06a67365b8fa5cc9020

domain/src/cost.rs
calebbourg: d88fcc74baeef7239af4a06a67365b8fa5cc9020

entity/src/provider.rs
calebbourg: That makes sense to me 520e265a0fdd289cf7c103448cffbe902144a55a

entity/src/cost_pricing_config.rs
calebbourg: Updated in 88ab842cfd27eafc7e052e41be9066b4195bdf19

entity/src/platform_cost_metrics.rs
calebbourg: Updated in Updated in 88ab842cfd27eafc7e052e41be9066b4195bdf19

migration/src/m20260529_000001_add_cost_tables.rs
calebbourg: Updated in 88ab842cfd27eafc7e052e41be9066b4195bdf19

domain/src/webhook/recording_done.rs
calebbourg: Moved this into a single spawn: 349dbc443198e03a36dbee995a99243869f310d6

migration/src/m20260529_000001_add_cost_tables.rs
calebbourg: Added fb04b51a5a97b940ed042087f2832140157b0844

migration/src/m20260529_000001_add_cost_tables.rs
calebbourg: Added in 23cabb3e08ca414496da095af4bcada14dbbd4b1

migration/src/m20260529_000001_add_cost_tables.rs
calebbourg: 23cabb3e08ca414496da095af4bcada14dbbd4b1

migration/src/m20260529_000001_add_cost_tables.rs
calebbourg: 23cabb3e08ca414496da095af4bcada14dbbd4b1

domain/src/webhook/transcript_done.rs
calebbourg: Comment added c8fa3153bce46c1027b22bc916c21b0b98e623dc

domain/src/cost.rs
calebbourg: 024c061c403a8aa68725d57147de8d2b0552f3f2

entity/src/cost_unit.rs
calebbourg: Updated 68ee8b6b21a889e5740f768e777cbeaa6645fa91

entity_api/src/platform_cost_metrics.rs
calebbourg: Didn't know this technique existed! 9a2ed4740d17cfaf5bc9c970c18dfdae153664cd

domain/src/cost.rs
```diff
@@ -0,0 +1,291 @@
...
+        CostMetricsModel {
+            id: Id::new_v4(),
+            provider: Provider::RecallAi,
+            metric,
+            coaching_session_id: Some(recording.coaching_session_id),
+            source_record_id,
+            cost_low: quantity * rate.cost_per_unit_low,
+            cost_high: quantity * rate.cost_per_unit_high,
+            cost_avg: quantity * rate.cost_per_unit_avg,
+            created_at: chrono::Utc::now().fixed_offset(),
```
jhodapp: These two fields are computed here but then discarded. `platform_cost_metrics::create` calls `into_active_model()` and unconditionally overrides both:

```rust
active_model.id = Set(Id::new_v4());                      // platform_cost_metrics.rs:14
active_model.created_at = Set(chrono::Utc::now().fixed_offset()); // :15
```

So the `id` and `created_at` you set here never reach the DB, and a reader has to cross-reference the CRUD layer to know which `now()` applies. It's not a bug, just a duplicated computation across the layer boundary.

Since `create` owns both fields, could we drop them here and let `create` be the single source? Either pass an `ActiveModel` with those two left unset, or have `create` take the non-generated fields explicitly.

domain/src/cost.rs
```diff
@@ -0,0 +1,291 @@
...
+        db,
+        CostMetricsModel {
+            id: Id::new_v4(),
+            provider: Provider::RecallAi,
+            metric,
+            coaching_session_id: Some(recording.coaching_session_id),
+            source_record_id,
+            cost_low: quantity * rate.cost_per_unit_low,
+            cost_high: quantity * rate.cost_per_unit_high,
+            cost_avg: quantity * rate.cost_per_unit_avg,
```
jhodapp: We pushed the seconds→unit conversion onto the `Unit` enum (`quantity_from_seconds`), which reads really well. The cost multiply is the mirror-image arithmetic but it stayed in the domain layer:

```rust
cost_low:  quantity * rate.cost_per_unit_low,
cost_high: quantity * rate.cost_per_unit_high,
cost_avg:  quantity * rate.cost_per_unit_avg,
```

For code symmetry we could give the rate `Model` the behavior too:

```rust
// entity/src/cost_pricing_config.rs
pub struct CostRange { pub low: Decimal, pub high: Decimal, pub avg: Decimal }

impl Model {
    pub fn cost_for(&self, quantity: Decimal) -> CostRange {
        CostRange {
            low:  quantity * self.cost_per_unit_low,
            high: quantity * self.cost_per_unit_high,
            avg:  quantity * self.cost_per_unit_avg,
        }
    }
}
```

Then the call site is just `let cost = rate.cost_for(quantity);`.
`

jhodapp (changes requested):
@calebbourg Just a couple more small pieces of code polish and this should be ready to merge and deploy.

domain/src/cost.rs
calebbourg: Good call b14bf534d1e41c8e7878e806cd0759a76ad6ab48

domain/src/cost.rs
calebbourg: Updated! 3aafd10b0a862948fb540f6689dd464af05759dd

jhodapp (approved):

PR #356, opened 2026-06-11
Coaching Session Series: first-class entity for recurring sessions

rnambaale:
## Description
Introduces `coaching_session_series` as a first-class entity, so coaches can create, read, reschedule, and delete a recurring series as a single unit instead of operating on N individual session rows.

## New endpoints

| Method | Path | Purpose |
|---|---|---|
| `POST`   | `/coaching_session_series`        | Create a series + materialize sessions in one txn |
| `GET`    | `/coaching_session_series`        | List series for a relationship (metadata only) |
| `GET`    | `/coaching_session_series/:id`    | Read one series + its linked sessions, sorted by date |
| `PUT`    | `/coaching_session_series/:id`    | Reschedule — replace the rule, re-materialize the future |
| `DELETE` | `/coaching_session_series/:id`    | Delete series + future sessions; past sessions survive |

The old `POST /coaching_sessions/recurring` route is removed.

## Data model

```
coaching_session_series
  id                       UUID PK
  coaching_relationship_id UUID FK -> coaching_relationships (CASCADE)
  rule                     JSONB  -- typed via domain::coaching_session_series::SeriesRule
  created_by_user_id       UUID FK -> users (RESTRICT)
  created_at / updated_at  TIMESTAMPTZ

coaching_sessions
  + coaching_session_series_id  UUID NULL  FK -> ... (ON DELETE SET NULL)
```

The FK uses `SET NULL` (not `CASCADE`) so application code drives the policy on what to do with past vs. future sessions when a series is removed. Past sessions become orphan one-offs with notes/Tiptap/meeting URLs intact.

The JSONB `rule` stores `{ start_at, recurrence, duration_minutes }`. Duration is resolved (via the coach's defaulting cascade) at series-create time and persisted on the rule, so a later default change can't silently flip the series's duration on reschedule.

## Behavior

**Create:** validates the rule, resolves duration, inserts the series row, bulk-inserts the materialized sessions linked via `coaching_session_series_id`. Lazy fields (`provider`, `collab_document_name`, `meeting_url`, `hydrated_at`) stay NULL on the new rows — they're populated by `coaching_session::ensure_hydrated` on first read, same as before.

**Reschedule + delete:**
- Future sessions are loaded inside a transaction.
- Each one's id is locked via `pg_advisory_xact_lock`, serializing against any in-flight `ensure_hydrated` on the same row.
- Future rows are bulk-deleted; notes / goal links / recordings / transcriptions cascade via existing FKs.
- For reschedule, the rule is rewritten and `bulk_create_recurring` runs with the new dates.
- After commit, captured Tiptap doc names are cleaned up **best-effort**. Failures are logged, not propagated.

### Testing Strategy
- [x] Unit / mock-DB tests for `entity_api::coaching_session_series` CRUD and `domain::coaching_session_series::{create_with_sessions, reschedule, delete_with_future_sessions}`.
- [x] Extractor integration tests covering coach/coachee/outsider paths against the three new extractors.
- [x] Controller AuthZ test for `POST /coaching_session_series` (coach-only inline check).
- [x] `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test --workspace` all pass.
- [ ] Manual E2E against the running stack: create → open one session (hydrates) → reschedule → confirm hydrated doc cleaned up in Tiptap → delete series → past sessions survive with `series_id = NULL`.

### Concerns
This PR introduces a breaking change when creating a recurring session. It would need to be deployed together with this [frontend PR](https://github.com/refactor-group/refactor-platform-fe/pull/420)


domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,581 @@
...
+                for name in &doc_names_to_cleanup {
+                    if let Err(err) = tiptap.delete(name).await {
+                        warn!(
+                            "Tiptap cleanup failed for orphaned doc {name:?} after series \
+                             delete {series_id}: {err}"
+                        );
+                    }
+                }
+            }
+            Err(err) => {
```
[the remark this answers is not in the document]
rnambaale: Good instinct on this one. I think the only code that attaches a session to an existing series is `bulk_create_recurring`, and the only caller that uses an existing `series_id` is `reschedule` (a one-off `coaching_session::create` always sets `series_id = None`; `create_with_sessions` uses a fresh id).

So there is a low probability of new sessions getting attached to the series we have selected here.

domain/src/error.rs
```diff
@@ -126,6 +126,17 @@ impl From<EntityApiError> for Error {
...
+impl From<serde_json::Error> for Error {
+    fn from(err: serde_json::Error) -> Self {
+        Error {
+            source: Some(Box::new(err)),
+            error_kind: DomainErrorKind::Internal(InternalErrorKind::Other(
+                "JSON serialization error".to_string(),
+            )),
+        }
+    }
+}
```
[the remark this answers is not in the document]
rnambaale: If are referring to Rule #2, that targets entity::* types skipping entity_api; `serde_json::Error` is external and skips no layer, so it doesn't apply. I think we need to keep this as is.

web/src/extractors/coaching_session_series_access.rs
```diff
@@ -0,0 +1,515 @@
...
+
+    fn test_series(
+        coaching_relationship_id: Id,
+        created_by_user_id: Id,
+    ) -> CoachingSessionSeriesApi::Model {
+        let now = Utc::now();
+        CoachingSessionSeriesApi::Model {
+            id: Id::new_v4(),
+            coaching_relationship_id,
+            rule: serde_json::json!({}),
```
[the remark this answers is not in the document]
rnambaale: Removed usage of `find_by_id_with_sessions` fn entirely. 025c0249925159b97bafb4d2707be8c0dea3e30a

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,610 @@
...
+                warn!(
+                    "Could not construct Tiptap client to clean up {} orphaned doc(s) \
+                     after series reschedule {series_id}: {err}",
+                    doc_names_to_cleanup.len()
+                );
+            }
+        }
+    }
```
jhodapp: The best-effort Tiptap cleanup here is duplicated verbatim in `delete_with_future_sessions` (lines 231–251) with only the `"reschedule"`/`"delete"` log word differs. I suggest extracting a private helper so the two log messages can't drift apart:

```rust
async fn cleanup_orphaned_docs(config: &Config, doc_names: &[String], series_id: Id, op: &str) {
  if doc_names.is_empty() {
      return;
  }
  match TiptapDocument::new(config).await {
      Ok(tiptap) => {
          for name in doc_names {
              if let Err(err) = tiptap.delete(name).await {
                  warn!("Tiptap cleanup failed for orphaned doc {name:?} after series {op} {series_id}: {err}");
              }
          }
      }
      Err(err) => warn!(
          "Could not construct Tiptap client to clean up {} orphaned doc(s) after series {op} {series_id}: {err}",
          doc_names.len()
      ),
  }
}
```

Both call sites then collapse to `cleanup_orphaned_docs(config, &doc_names_to_cleanup, series_id, "reschedule").await;` and the `if !is_empty()` guard moves inside.
`

entity_api/src/coaching_session.rs
```diff
@@ -153,6 +158,48 @@ pub async fn find_by_id(db: &impl ConnectionTrait, id: Id) -> Result<Model, Erro
...
+
+/// Bulk-deletes coaching sessions by id. Returns the number of rows removed.
+/// FK cascades on dependent rows (goals, recordings, transcriptions) handle
+/// child cleanup.
+pub async fn bulk_delete_by_ids(db: &impl ConnectionTrait, ids: &[Id]) -> Result<u64, Error> {
+    if ids.is_empty() {
+        return Ok(0);
+    }
+    let result = Entity::delete_many()
+        .filter(Column::Id.is_in(ids.to_vec()))
```
jhodapp: `is_in(ids.to_vec())` clones the whole slice. `Id` is `Copy` and `is_in` takes `IntoIterator<Item = V>`, so an iterator of values avoids the unnecessary allocation:

```rust
.filter(Column::Id.is_in(ids.iter().copied()))
```

This also matches the pattern already used elsewhere in this same file (e.g. lines 741, 759, 777).
`

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,610 @@
...
+/// Typed shape of the JSONB `rule` column on `coaching_session_series`.
+/// Stored at create time and re-read at reschedule time, which is why
+/// `duration_minutes` is captured here — a coach who updates their default
+/// duration between create and reschedule shouldn't see the existing series
+/// flip durations under them.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SeriesRule {
+    pub start_at: NaiveDateTime,
+    pub recurrence: Recurrence,
+    pub duration_minutes: i16,
```
jhodapp, line 34: The `SeriesRule` doc comment says `duration_minutes` is persisted so that *"a coach who updates their default duration between create and reschedule shouldn't see the existing series flip durations under them."* But `reschedule` never reads the stored rule — it takes a fresh `new_requested_duration` and runs it through `resolve_duration` (line 129), which falls back to the coach's **current** default when the request omits a duration:

```rust
pub async fn resolve_duration(db, coach_id, requested) -> Result<Duration, Error> {
    if let Some(d) = requested { return Ok(d); }
    let coach = crate::user::find_by_id(db, coach_id).await?;
    Ok(Duration::from_minutes_unchecked(coach.default_coaching_session_duration_minutes))
}
```

So a reschedule with `duration_minutes: None` *does* flip the series to the coach's new default — exactly what the comment claims is prevented. The persisted value is also never deserialized on the backend, so the `Deserialize` derive on `SeriesRule` is currently dead here.

Pick one:
- **If the invariant is intended:** on reschedule, when the request omits a duration, read the old rule (`serde_json::from_value::<SeriesRule>(series.rule)?`) and reuse its `duration_minutes` instead of calling `resolve_duration`. That makes both the derive and the comment true.
- **If the stored duration is purely for API consumers:** soften the comment so it doesn't claim a backend guarantee the backend doesn't enforce.
`

web/src/extractors/coaching_session_series_access.rs
```diff
@@ -0,0 +1,511 @@
...
+/// Shared lookup: extract `:id` from the path, load the series, load its
+/// parent relationship, and pull the authenticated user. The caller decides
+/// what membership rule to enforce.
+async fn resolve<S>(
+    parts: &mut Parts,
+    state: &S,
+) -> Result<
+    (
+        CoachingSessionSeriesApi::Model,
+        domain::coaching_relationships::Model,
```
jhodapp: `coaching_relationships` is already imported at the top of the file, but the `resolve` return tuple writes the fully-qualified form:

```rust
domain::coaching_relationships::Model,   // imported as `coaching_relationships`
domain::users::Model,                    // `users` not imported
```

Use the short name (`coaching_relationships::Model`) for consistency, or drop the unused import. Cosmetic, no behavior change.
`

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,610 @@
...
+use service::config::Config;
+
+pub use coaching_session::{Frequency, Recurrence, RecurrenceError};
+pub use entity::coaching_session_series::Model;
+pub use entity_api::coaching_session_series::find_by_id;
+
+/// Typed shape of the JSONB `rule` column on `coaching_session_series`.
+/// Stored at create time and re-read at reschedule time, which is why
+/// `duration_minutes` is captured here — a coach who updates their default
+/// duration between create and reschedule shouldn't see the existing series
```
jhodapp, line 28: **Severity: High — behavioral defect**

The `SeriesRule` doc promises a coach *"shouldn't see the existing series flip durations under them"* between create and reschedule. But `reschedule` never reads the stored rule — it re-resolves from scratch at line 129:

```rust
let resolved_duration =
    entity_api::coaching_session::resolve_duration(db, coach_id, new_requested_duration).await?;
```

and `resolve_duration` falls back to the coach's **current** default when the request omits a duration:

```rust
if let Some(d) = requested { return Ok(d); }
let coach = crate::user::find_by_id(db, coach_id).await?;
Ok(Duration::from_minutes_unchecked(coach.default_coaching_session_duration_minutes))
```

So a reschedule with `duration_minutes: None` re-materializes every future session at the coach's *new* default — exactly the flip the doc says is prevented. The persisted `duration_minutes` is also never deserialized backend-side, which makes the `Deserialize` derive on `SeriesRule` (line 30) dead code.

**Fix — pick one:**
- If the invariant holds, reuse the stored duration when the request omits one:
  ```rust
  let resolved_duration = match new_requested_duration {
      Some(d) => entity_api::coaching_session::resolve_duration(db, coach_id, Some(d)).await?,
      None => {
          let existing = coaching_session_series::find_by_id(db, series_id).await?;
          let prior: SeriesRule = serde_json::from_value(existing.rule)?;
          Duration::from_minutes_unchecked(prior.duration_minutes)
      }
  };
  ```
  (this also makes the `Deserialize` derive live)
- If clients are expected to send a duration on every reschedule, delete the misleading sentence and the field's stated rationale.
`

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,610 @@
...
+use crate::Id;
+use chrono::NaiveDateTime;
+use entity_api::coaching_session_series;
+use log::warn;
+use sea_orm::{DatabaseConnection, TransactionTrait};
+use serde::{Deserialize, Serialize};
+use service::config::Config;
+
+pub use coaching_session::{Frequency, Recurrence, RecurrenceError};
+pub use entity::coaching_session_series::Model;
```
jhodapp: **Severity: Medium — layer-boundary violation (coding-standards.md "Type Re-Export Boundary")**

```rust
pub use entity::coaching_session_series::Model;   // reaches past entity_api into entity
```

The standards name this exactly: *"Adding a `pub use entity::<module>::<Type>;` re-export inside a domain module file … reaches across the entity_api boundary … Re-exports of entity types belong at lib.rs only."* `coaching_session_series` is in neither `pub use entity::{…}` (entity_api/lib.rs) nor `pub use entity_api::{…}` (domain/lib.rs).

Root cause is a naming collision the rest of the codebase avoids by convention (entity tables plural `coaching_sessions`, domain concepts singular `coaching_session`). This feature used the same name for both the entity table module and the domain orchestration module, so it couldn't go in the lib.rs `pub use` blocks without colliding with `pub mod coaching_session_series`.

**Fix — route the type one hop per layer (no lib.rs change, so no collision):**
```rust
// entity_api/src/coaching_session_series.rs  (entity_api re-exports the entity Model — allowed within its layer)
pub use entity::coaching_session_series::Model;

// domain/src/coaching_session_series.rs:22   (domain re-exports from entity_api, not entity)
pub use entity_api::coaching_session_series::Model;
```
Call-site path `domain::coaching_session_series::Model` is unchanged.
`

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,610 @@
...
+        coach_id,
+        updated_series.id,
+        new_dates,
+        Some(resolved_duration),
+    )
+    .await?;
+
+    txn.commit().await.map_err(entity_api::error::Error::from)?;
+
+    if !doc_names_to_cleanup.is_empty() {
```
jhodapp: **Severity: Medium — untested critical path**

Both `reschedule` (lines 172–192) and `delete_with_future_sessions` (lines 231–251) end with a best-effort Tiptap cleanup branch — the only externally-observable side effect of these operations. Every domain test deliberately skips it: the reschedule test even comments *"Tiptap is never contacted because no future session has a collab doc,"* and all test sessions use `collab_document_name: None`.

Because errors there are swallowed, a defect leaks documents instead of failing the request — the kind of silent regression that survives refactoring unnoticed, which is exactly why it needs a test.

**Fix:** — extract the cleanup into one helper, then add a single test with a future session carrying `collab_document_name: Some(...)`. The existing `test_config()` points Tiptap at `127.0.0.1:0`, so the delete fails-and-warns, exercising the swallow-and-continue path and asserting it doesn't propagate.
`

web/src/controller/coaching_session_series_controller.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    let db = app_state.db_conn_ref();
+
+    // Inline coach-only AuthZ — body-bearing routes parse the body in the
+    // controller (matching the existing create_recurring pattern).
+    let relationship =
+        domain::coaching_relationship::find_by_id(db, params.coaching_relationship_id).await?;
+    if relationship.coach_id != user.id {
+        return Err(Error::Web(crate::error::WebErrorKind::Auth));
+    }
```
jhodapp: **Severity: Medium — split authz idiom (coding-standards.md "Thin Controllers")**

`create` hand-rolls authorization in the handler (lines 58–61).

while `read`/`index`/`update`/`delete` enforce access through `FromRequestParts` extractors. The constraint is real — the relationship id is in the JSON body, and extractors run before body parsing — but that splits "who's allowed" across two idioms (extractor for 4 routes, controller for 1).

**Fix:** push the coach check into the domain entry point `create_with_sessions`, which already receives the body's data. The controller then calls one function and maps one error, like the other handlers, and the authz rule lives next to the operation it guards.
`

web/src/controller/coaching_session_series_controller.rs
```diff
@@ -0,0 +1,324 @@
...
+) -> Result<impl IntoResponse, Error> {
+    debug!(
+        "PUT reschedule coaching_session_series id={} params={params:?}",
+        series.id
+    );
+
+    let db = app_state.db_conn_ref();
+    let relationship =
+        domain::coaching_relationship::find_by_id(db, series.coaching_relationship_id).await?;
+    let requested_duration = CoachingSessionApi::parse_duration_minutes(params.duration_minutes)?;
```
jhodapp, line 162: **Severity: Low — redundant query**

`CoachingSessionSeriesCoachAccess` already loads the series *and* its relationship inside `resolve()` to prove `relationship.coach_id == user.id`, but returns only the series. The handler then re-loads the relationship purely to read `coach_id` (line 171):

```rust
CoachingSessionSeriesCoachAccess(series): CoachingSessionSeriesCoachAccess,
// ...
let relationship =
    domain::coaching_relationship::find_by_id(db, series.coaching_relationship_id).await?;
```

That's an extra round-trip per reschedule for data the extractor had in hand. `read`/`index` don't have this issue.

**Recommendation — either:**
- Surface it from the gate: `CoachingSessionSeriesCoachAccess(pub Model, pub Id /* coach_id */)`, or
- Since the gate guarantees the caller *is* the coach, add `AuthenticatedUser(user)` to `update` and pass `user.id` as the coach id — no second query at all.
`

jhodapp (changes requested):
Thanks for this @rnambaale, I ran several different reviews with Claude's help. Let me know if you have any questions about the suggestions.

migration/src/m20260611_000000_add_coaching_session_series.rs
jhodapp, line 1: @rnambaale Also a reminder to update docs/db/refactor_platform_rs.dbml and regenerate docs/db/refactor_platform_rs.png using dbdiagram.io.

domain/src/coaching_session_series.rs
rnambaale: Done. dada2fce8edd6b01aff8ee6aa82752a0c48af691

entity_api/src/coaching_session.rs
rnambaale: Resolved. 6e7cc5c6e57519bf6013413e75fd81d4b2a5c8ea

domain/src/coaching_session_series.rs
rnambaale: Nice find. 

I have opted for the latter, I think it is better to have matching duration on the existing and new sessions by default.

35924997c8c9aae9cb4fc3a3663bd25285767aa9

domain/src/coaching_session_series.rs
rnambaale: Fixed in 35924997c8c9aae9cb4fc3a3663bd25285767aa9.

We opted to read the old series rule.

domain/src/coaching_session_series.rs
rnambaale: Fixed: aaa1bc5514a5ba9035cef8f95dde996a0878d6c9

domain/src/coaching_session_series.rs
rnambaale: Added fc3baba7402601d6f8c33dea01d2d068cf2f177f

web/src/controller/coaching_session_series_controller.rs
rnambaale: Fixed in: 2779397054982f1e4db86c142b01d95990b982f6

While making this refactor, I wondered whether 403 would actually be a more appropriate status code for cases where the `logged in user` != `relationship coach`. I have left it as `401` for now since that seems to be the convention.

Happy to revisit of you think that is necessary.

web/src/controller/coaching_session_series_controller.rs
rnambaale: Fixed. 4522b9d5559656c8b1e670dd216a666a48664f0f

migration/src/m20260611_000000_add_coaching_session_series.rs
rnambaale: I have updated it here: 270363e0b77482180917ed4a1d1b24b6272d0727

domain/src/coaching_session_series.rs
```diff
@@ -0,0 +1,581 @@
...
+            id: Id::new_v4(),
+            coaching_relationship_id: relationship_id,
+            coaching_session_series_id: Some(series.id),
+            collab_document_name: None,
+            date,
+            duration_minutes: 60,
+            meeting_url: None,
+            provider: None,
+            created_at: now.into(),
+            updated_at: now.into(),
```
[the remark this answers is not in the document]
jhodapp: @rnambaale don't forget to have a look at this issue flagged by Greptile. Thanks.

domain/src/coaching_session_series.rs
jhodapp: Agreed, thanks.

domain/src/coaching_session_series.rs
jhodapp: @rnambaale meaning you're reusing the existing session duration and ignoring any new duration value sent along in the endpoint parameters?

web/src/controller/coaching_session_series_controller.rs
jhodapp: @rnambaale hmm yeah, and 403 does seem to be more appropriate here. But I don't think it makes sense to deviate from the convention currently and we should revisit this holistically. Would you mind filing an issue describing this situation as a refactoring task? Thanks for flagging this.

web/src/router.rs
```diff
@@ -281,10 +289,6 @@ pub fn coaching_sessions_routes(app_state: AppState) -> Router {
             "/coaching_sessions",
             post(coaching_session_controller::create),
         )
-        .route(
-            "/coaching_sessions/recurring",
-            post(coaching_session_controller::create_recurring),
-        )
         .merge(
             // Get /coaching_sessions
```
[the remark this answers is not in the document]
jhodapp: @rnambaale something to be aware of and double check.

jhodapp (changes requested):
A few more things to have a look at and double check.

domain/src/coaching_session_series.rs
rnambaale: @jhodapp

Not quite. On create, an explicit `duration_minutes` in the request always takes precedence; the coach's stored default is only used when the field is omitted. 

On reschedule, we reuse the series rule's stored duration as the default, but that is overridden if an explicit `duration_minutes` is requested. We never read from the coach's currently stored default.

web/src/controller/coaching_session_series_controller.rs
rnambaale: Sure. I have created the issue here: https://github.com/refactor-group/refactor-platform-rs/issues/358

web/src/router.rs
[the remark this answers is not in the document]
rnambaale: Right, so the plan is to deploy this together with this frontend [PR](https://github.com/refactor-group/refactor-platform-fe/pull/420).

I've added that as a concern in the PR description.

domain/src/coaching_session_series.rs
jhodapp: @rnambaale perfect, that seems like the most logical way of solving it. Thanks.

web/src/router.rs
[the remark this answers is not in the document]
jhodapp: Thanks.

jhodapp (approved):
Looks great, thank you Raymond.

PR #367, opened 2026-06-27
Return 403 for authorization failures instead of 401

### LINKED ISSUE #358: Use 403 Forbidden for authorization failures instead of 401 Unauthorized
## Summary

We currently return **401 Unauthorized** for authorization failures (an authenticated user attempting an action they're not permitted to perform).

## Current behavior

`EntityErrorKind::Unauthenticated` is the catch-all for both authentication and authorization failures, and it maps to 401:

```rust
// web/src/error.rs:108
EntityErrorKind::Unauthenticated => {
    warn!("EntityErrorKind::Unauthenticated: Responding with 401 Unauthorized. Error: {self:?}");
    (StatusCode::UNAUTHORIZED, "UNAUTHORIZED").into_response()
}
```

### Example trigger (the review site)

In `domain/src/coaching_session_series.rs:63-69`, an authenticated user who is *not the coach* on the target relationship is rejected with `Unauthenticated` → 401, where 403 would be correct:

```rust
if relationship.coach_id != acting_user_id {
    return Err(Error {
        source: None,
        error_kind: DomainErrorKind::Internal(InternalErrorKind::Entity(
            EntityErrorKind::Unauthenticated,
        )),
    });
}
```

### Other affected sites (non-exhaustive)

- The access extractors return a bare `(StatusCode::UNAUTHORIZED, ...)` for authz failures — e.g. `web/src/extractors/coaching_session_series_access.rs` (coach-only and participant gates), and the analogous existing extractors (`coaching_session_access.rs`, etc.).
- Any domain authz check funneling through `EntityErrorKind::Unauthenticated`.

Genuine *authentication* failures (no/invalid session) — e.g. `entity_api/src/user.rs`, `magic_link_token`, `user_session_controller` login — should remain **401**.

## Why holistic

The 401/403 distinction is cross-cutting. A correct fix likely needs:

- A distinct error kind for authorization (e.g. `EntityErrorKind::Forbidden`) separate from `Unauthenticated`, mapped to 403 in `web/src/error.rs`.
- An audit of every `Unauthenticated` site and every bare `StatusCode::UNAUTHORIZED` in extractors to classify it as authn (401) vs authz (403).
- Coordination with the frontend, which may branch on status codes (a 401 might trigger a re-login/redirect that a 403 should not).

## Acceptance criteria

- [ ] Authorization failures (authenticated but not permitted) return **403**.
- [ ] Authentication failures (missing/invalid credentials) continue to return **401**.
- [ ] All current `Unauthenticated` usages and bare-401 extractor rejections are audited and classified.
- [ ] Frontend handling of 401 vs 403 is confirmed/updated.

rnambaale:
## Description
Authorization failures (an authenticated user attempting an action they're not
permitted to perform) currently return **401 Unauthorized**. The correct status is
**403 Forbidden** — 401 means "who are you?" (missing/invalid credentials), 403 means
"I know who you are, and you can't do this." Conflating them is wrong per HTTP semantics
and harmful on the frontend, where a 401 typically triggers a re-login/redirect that a
403 should not.

This change introduces a distinct authorization error kind, maps it to 403 in the single
web error switch, and audits every authn-vs-authz site so authentication failures keep
returning 401 while authorization failures now return 403.

### Changes
* Added `EntityErrorKind::Forbidden` (authorization) in `domain/src/error.rs`, documented
  as distinct from `Unauthenticated` (authentication), and mapped it to **403** in
  `web/src/error.rs` (`Unauthenticated` stays **401**).
* Domain authz site `domain/src/coaching_session_series.rs` (non-coach creating a series)
  now yields `Forbidden` instead of `Unauthenticated`.
* Flipped authenticated-but-not-permitted gates from **401 → 403** across all access
  extractors (`coaching_session_series_access`, `coaching_session_access`,
  `coaching_relationship_access`, `organization_member_access`) and the `protect/*`
  middleware (`coaching_sessions`, `jwt`, `notes`, `actions`, `agreements`, `goals`, and
  the `users/*` self-ownership gates).
* Kept **401** for genuine authentication failures: the `AuthenticatedUser` extractor and
  `require_auth` (no/invalid session), webhook signature verification (`svix_signature`),
  and the login / magic-link / token-validation paths.
* Updated the OpenAPI response docs on the coaching-session-series endpoints to include
  403, and updated/added tests (renamed `*_401_*` authz tests to `*_403_*`, added a
  participant-denial 403 test for `coaching_session_access`, and a `web::error` unit test
  locking the `Forbidden → 403` / `Unauthenticated → 401` mapping).

### Testing Strategy
* `cargo build`, `cargo clippy --all-targets --features mock`, and `cargo fmt --check` are
  clean.
* `cargo test --features mock` passes (all suites green).
* Manual contract check:
  * Authenticated non-coach calling a coach-only series route → **403**.
  * Authenticated non-participant reading a session/relationship → **403**.
  * Unauthenticated request to any protected route → **401**.
  * Invalid/missing webhook signature → **401**.

### Concerns
* **Frontend coordination required (separate repo):** the 401/403 split is now meaningful.
  Frontend should treat **401** as "re-authenticate / redirect to login" and **403** as
  "not permitted — surface a message, do NOT redirect."


web/src/protect/users/coaching_sessions.rs
```diff
@@ -24,6 +24,6 @@ pub(crate) async fn index(
             "Unauthorized: user_id {} does not match authenticated_user_id {}",
             user_id, authenticated_user.id
         );
-        (StatusCode::UNAUTHORIZED, "Unauthorized").into_response()
+        (StatusCode::FORBIDDEN, "Forbidden").into_response()
```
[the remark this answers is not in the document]
rnambaale: 237d9997a217186ac30374a0c1569de8a0c33ebf

web/src/protect/coaching_sessions.rs
jhodapp: Should this also be changed to FORBIDDEN?

jhodapp: Should this also be changed to FORBIDDEN?

web/src/protect/users/mod.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) async fn read(
             "Unauthorized: user_id {} does not match authenticated_user_id {}",
             user_id, authenticated_user.id
         );
```
jhodapp, line 32: Don't forget to update the error wording here to forbidden instead of unauthorized. I recommend checking all of the error statements around each line you changed as well.

jhodapp (changes requested):
Just a couple of changes needed, thanks Raymond.

web/src/protect/users/mod.rs
rnambaale: 9bf0412c33ebd7a81c8bdd8d6983f6d327289130

web/src/protect/coaching_sessions.rs
rnambaale: I thought this might be a slightly closer to 404, than 403. The error branch is matched when a session does not exist in the DB. 

I happy to switch to either of those depending on the convention we want to use here.

jhodapp: @rnambaale looking at this again more deeply, we are also assuming here that the only kind of error that `find_by_id()` will return is related to authorization, which isn't always true (it looks like it could also be a `EntityApiErrorKind::RecordNotFound`), and perhaps this is your point why you think 404 makes the most sense in this context?

jhodapp: @rnambaale and maybe the same here, maybe 404 when the relationship can't be found is the real error to return here.

rnambaale: @jhodapp 

Yes, so I think there are two kinds of errors `find_by_id()` can return: `EntityApiErrorKind::RecordNotFound`, and a `EntityApiErrorKind::SystemError`

Rather asking web to convert entity_api errors, we opted to use the existing `coaching_session::find_by_id_with_coaching_relationship` domain fn that returns both the session + relationship at once.

A missing session maps to 404, and a DB/system failure maps to 503. Now 403 only comes last.

0a163f4a14f0597cdb18482b06df41d8558e26ee

rnambaale: 0a163f4a14f0597cdb18482b06df41d8558e26ee

jhodapp: Sounds great, thanks for making this fix and raising this concern.

jhodapp (approved):

Issue #378, opened 2026-08-08
Collab JWT outlives organization removal by up to 24h


jhodapp:
Follow-up to #377, which made removing a member from an organization revoke their access to that organization's coaching data.

That work closes the request/response paths and the SSE event stream. It does **not** close the collaborative document editor, because the collab token is a bearer JWT that is validated independently of the API.

## The gap

`domain/src/jwt/mod.rs` mints collab tokens with `exp: now + 24 hours`. The collab server validates the signature and expiry on its own; it never asks the API whether the bearer is still a member.

So the sequence is:

1. A member opens a coaching session and the frontend requests a collab token.
2. An admin removes them from the organization.
3. `GET /jwt/generate_collab_token` now correctly returns 403, so they cannot get a *new* token.
4. The token they already hold keeps working against the collab server until it expires.

During that window the removed user can still read and edit the session's collaborative document. Everything else is already denied.

## Why it was left out of #377

There is no revocation path today. Fixing it is a design change rather than a predicate change, and #377 was deliberately scoped to "removing a member revokes their access" through the API surfaces that consult the database on every request.

## Options

- **Short TTL plus refresh.** Cut `exp` to a few minutes and have the frontend re-request. Revocation then takes effect within one TTL, and the refresh call goes through the already-fixed 403 path. Costs a periodic request per open document.
- **Membership check on the collab side.** The collab server asks the API (or reads the database) before accepting a connection, and periodically thereafter. Strongest, but couples the collab server to membership state.
- **A revocation list.** Publish removals to a short-lived denylist the collab server consults. Adds moving parts.

Short TTL plus refresh is the smallest change that meaningfully shrinks the window.

## Acceptance

- A user removed from an organization loses collaborative document access within a bounded, documented window.
- The window is short enough to state plainly in the admin UI, or removal is immediate.
- The still-member coach's editing is uninterrupted, including across a token refresh.

PR #379, opened 2026-08-08
Remove from organization: revoke the member's access, keep the coach's history

### LINKED ISSUE #377: Remove from organization: revoke the member's access but keep the coach's history
Follow-up to #376, which added `DELETE /organizations/{organization_id}/users/{user_id}/role`.

That route currently **refuses with a 409** (`user_has_coaching_history`) when the member has any coaching sessions in the organization, because `coaching_sessions` references `coaching_relationships` with `NO ACTION` and deleting the relationship would fail, while `goals` and `coaching_session_series` would cascade away for the coach.

The 409 is safe but too strict in practice: any member who has actually been coached has session history, so removal is blocked for essentially every real member.

## Wanted behavior

Removing a member should **succeed** and:

- Delete only the `user_roles` row. Keep `coaching_relationships`, `coaching_sessions`, `notes`, `actions`, `agreements`, `goals`, and topics.
- **Revoke the removed user's access** to that organization's data.
- **Preserve the coach's access** to the same sessions, notes, and actions.

## Why this needs an authorization change, not just a delete

Every read path authorizes on relationship participation alone:

```rust
if coaching_relationship.coach_id == user.id
   || coaching_relationship.coachee_id == user.id
```

There is no `user_roles` check anywhere in `web/src/protect/`. So dropping the `user_roles` row alone hides the organization from the removed user's switcher (`organization::find_by_user` reads `user_roles`) **but does not revoke API access**. They keep full access to their sessions, notes, and actions with a direct URL or an API client.

The rule needs to become **participation AND current membership**, which denies the removed user while leaving the still-member coach untouched. This also closes a pre-existing gap: today anyone who was ever in a relationship retains access regardless of membership.

Sites doing the participation check:

- `web/src/protect/actions.rs`
- `web/src/protect/notes.rs`
- `web/src/protect/coaching_sessions.rs`
- `web/src/protect/agreements.rs`
- `web/src/protect/jwt.rs`
- `web/src/extractors/coaching_session_access.rs`
- `web/src/extractors/coaching_session_series_access.rs`
- `web/src/extractors/coaching_session_topic_access.rs`

## Re-adding a removed member

There is a unique index `coaching_relationships_coach_coachee_org` on `(coach_id, coachee_id, organization_id)`. `coaching_relationship::create` pre-checks it and returns a validation conflict. So re-adding a member **with their original coach pre-assigned** fails with "Coaching relationship already exists for this coach and coachee in the organization."

Re-add should reuse the surviving relationship instead of erroring, restoring access without duplicating data.

## Scope

- Remove the `user_has_coaching_history` guard and its `EntityApiErrorKind` / `EntityErrorKind` variants and the 409 mapping in `web/src/error.rs`.
- Keep the last-admin guard (with its `FOR UPDATE` lock) unchanged.
- Add membership to the authorization rule at the sites above.
- Make attach reuse an existing relationship rather than conflict.
- Decide whether a coach can still be removed while they have active coachees, since that leaves those coachees with a coach who is not a member.

## Acceptance

- Removing a member with session history returns 204 and deletes exactly one `user_roles` row.
- The removed user gets 403 on that organization's sessions, notes, and actions.
- The coach still reads the same sessions, notes, and actions.
- Re-adding the member with their original coach succeeds and restores access to the existing relationship.
- Coverage must include a live-database case; mocks cannot reach the foreign key or the cross-user authorization interaction.

jhodapp:
## Description

Removing a member from an organization now succeeds and actually revokes their access, without deleting any of their coaching history.

Two problems are fixed together. Removal used to refuse with a 409 whenever the member had any coaching sessions, which in practice is every real member. And dropping the membership row only hid the organization from their switcher: it left full API access intact, so a removed user kept reading their sessions, notes and actions with a direct URL or an API client.

### Changes

* **Removal is a single leaf delete.** `remove_from_organization` drops one `user_roles` row and stops. Coaching relationships, sessions, notes, actions, agreements, goals and topics all survive. The old code also deleted the member's coaching relationships, which is what forced the 409 in the first place: `coaching_sessions` references `coaching_relationships` with `NO ACTION`. Nothing references `user_roles(id)`, so the remaining delete has no dependents.
* **Authorization now requires membership, not just participation.** Adds `coaching_relationships::Model::grants_access_to`, and points 17 call sites at it across `web/src/protect/` and `web/src/extractors/`. The rule changes from "are you the coach or coachee" to "are you the coach or coachee **and** still a member of that relationship's organization". This also closes a pre-existing gap: anyone ever in a relationship previously kept access forever. Costs no extra query, since `user.roles` is already hydrated on every authenticated request.
* **Coach-only routes keep their narrower rule.** The four coach-only sites use `coach_id == user.id && grants_access_to(&user)`, so they gain membership without widening to the coachee.
* **SSE notify sets are membership-filtered.** Event recipients came from the surviving relationship, so a removed user holding an open stream kept receiving note, action, goal and session content. Adds `retain_organization_members` and routes all six notify-set builders through it.
* **Re-adding reuses the surviving relationship.** Because the relationship now outlives removal, re-adding a member with their original coach used to collide with the `coaching_relationships_coach_coachee_org` unique index. `coaching_relationship::create` now returns the existing relationship instead of erroring. This makes `POST /organizations/{id}/coaching_relationships` idempotent, which also fixes the same wall reached via the members UI "Assign coach" action.
* **Relationship-create authorization moved to the extractor pattern.** Replaces a `Predicate`/`UserIsAdmin` route layer plus a redundant `OrganizationMemberAccess` with a single `OrganizationAdminAccess`. Who may call it is unchanged; it drops two queries per request.
* **Removes the now-unreachable `UserHasCoachingHistory` error variant** and its 409 mapping across `entity_api`, `domain` and `web`, plus two dead `entity_api` helpers.

### Testing Strategy

**Automated.** 726 mock-based tests pass (entity 48, entity_api 282, domain 238, web 158), plus `cargo clippy`, `cargo fmt --check` and `cargo check --workspace` clean. Run the mock suite with:

```
cargo test -p entity_api -p domain -p web --features "domain/mock,web/mock"
```

New frozen test files cover the predicate truth table, revocation through the real router and session stack, the coach-only anti-widening case, membership filtering of notify sets, and relationship reuse. Every frozen assertion was mutation-checked: the guarded behavior was deliberately broken to confirm the test fails.

**Manual.** `docs/test-plans/remove_org_member_revokes_access_manual_testing.md` is committed and was executed against a real database via curl. Scenario C is the one that proves the feature and must be run with an API client, since the UI hiding an organization is not evidence of revocation. Verified results:

| Surface | Baseline (member) | After removal | Coach |
|---|---|---|---|
| coaching session, notes, actions, agreements, topics, goals, collab token | 200 | **403** | 200 |

Removal of a member with real session history succeeded, the coach kept every surface, re-adding with the original coach returned 201 and preserved the **original relationship id** with no duplicate, and the database was restored entirely through the API.

### Concerns

* **Re-add requires a super admin, by design.** Removal drops the shared membership that made the member visible to the org admin, so that admin's lookup returns nothing and attach returns 404. Confirmed intended. Scenario I in the test plan covers both halves, and the frontend has been told the acceptance criterion needs the super-admin qualifier.
* **Collab JWTs outlive revocation by up to 24h** (#378). A removed user cannot mint a new token, but one already in hand keeps working against the collab server until it expires. No revocation path exists today; fixing it is a design change.
* **`DELETE .../role` returns HTTP 200 with `{"status_code":204}`**, per this codebase's `ApiResponse` convention. Pre-existing, not changed here, but worth knowing if you assert on the transport status.
* **Foreign-key safety has no automated test.** There is no live-database harness in this repo, so it rests on the schema argument plus the manual run above.
* **One separate authorization gap** was found while verifying and is being handled outside this PR. It predates this work and is unrelated to it.
* **Coordinated with the frontend half** (refactor-platform-fe#435). No coordinated deploy is strictly required, but shipping together avoids the frontend carrying dead 409 handling.


jhodapp:
Agreed, and fixed in 1f0f825d. Verified every link in the chain first:

- The composite unique index `coaching_relationships_coach_coachee_org` exists on `(coach_id, coachee_id, organization_id)`.
- Line 109 was a bare `.insert(db).await?` with no conflict handling.
- A unique violation falls through `From<DbErr>` to the `_ =>` arm, so it surfaced as `SystemError`, not the existing relationship.
- In `attach_to_organization` the `?` propagates before `txn.commit()`, so the membership insert rolled back too. The admin got a 500 **and** the member was not added.

**One correction to the obvious fix.** Catching the violation and re-selecting would not work: in Postgres any error inside a transaction aborts it, so the recovery read would itself fail with `25P02 current transaction is aborted`. Since `create` receives `&txn` here, that path is unavailable.

The fix inserts `ON CONFLICT (coach_id, coachee_id, organization_id) DO NOTHING` and reads the winner back when no row is written. `DO NOTHING` succeeds having inserted nothing, so the transaction stays healthy and the recovery read can run. It also blocks correctly against a concurrent uncommitted insert: if that transaction commits we conflict and read its row, if it rolls back we insert normally.

Two API details worth recording, both of which would have made this silently not work:

- `on_conflict_do_nothing()` resolves to the **primary key** (`ON CONFLICT ("id")`), so it would not have caught this conflict at all. The conflict target has to name the index columns explicitly.
- For a returning insert on Postgres, the zero-row case surfaces as `DbErr::RecordNotFound`, not `RecordNotInserted`. `TryInsertResult` only maps the latter, so it never reports `Conflicted` here. Matching the wrong variant compiles and does nothing.

Generated statement matches the index exactly, same columns in the same order:

```sql
... ON CONFLICT ("coach_id", "coachee_id", "organization_id") DO NOTHING RETURNING ...
```

Covered by a new frozen test, `returns_the_winner_when_a_concurrent_insert_wins_the_race`, mutation-checked two ways: removing the recovery arm fails it, and dropping the `ON CONFLICT` clause fails it.

One thing your comment prompted that is worth flagging separately: my first attempt also moved the pair match from Rust into SQL as a tidy-up. That broke `does_not_reuse_a_reversed_pair`, because `MockDatabase` does not apply a `WHERE`, so the test could no longer tell a matching pair from a reversed one. Production behaviour was still correct, but the guard protecting directionality had been silently disarmed. Reverted, and the helper now carries a comment saying why the match stays in Rust.

jhodapp:
Thanks — all five verified real. Three fixed here in `bad54373`, one filed, one deliberately not filed.

**#2 notify-set duplication — fixed.** This was the strongest one, and the duplication was mine from the SSE phase rather than pre-existing. All six sites now call `coaching_relationship::notify_member_ids(db, &relationship)`, so the membership rule has one home. Verified every site had a relationship model in hand, so the helper takes `&Model` and reads as a sentence at each call.

**#5 inline path — fixed.** `use crate::users;` hoisted, signature is `&users::Model`, matching the `use crate::roles::Role;` right above it.

**#4 201 on reuse — fixed, and you were right that it matters.** The envelope carries `201` even when nothing was created, so it does not distinguish a reuse from a create. The OpenAPI description now says so explicitly, since a client could reasonably branch on it.

**#1 organization scan — comment fixed, optimization filed as #382.** You were right to push back on the rationale. Justifying production shape by "so mocks can observe it" is backwards, and that sentence is gone.

I did not make the query change in this PR, for one reason worth stating: `does_not_reuse_a_reversed_pair` currently proves that a swapped coach/coachee is not treated as the same relationship, by feeding a reversed row through the mock and watching it be rejected. Move the filter into SQL and `MockDatabase` stops applying the `WHERE`, so the test can no longer tell the two cases apart. Your suggested trade — assert on the generated SQL instead — is reasonable, but it swaps a proof about behaviour for a proof about a query string, on a property where getting it wrong attaches the wrong pair. That felt worth a deliberate decision rather than a drive-by in an already-large PR, so #382 carries the full context including the test trade-off.

Worth noting I hit exactly this failure mode while fixing the race here: my first attempt moved the match into SQL as a tidy-up, and it silently disarmed that same test. Production stayed correct, the guard did not.

**#3 organization-existence leak — real, not filed.** Confirmed: the extractor resolves the organization and 404s before evaluating the role, so 403 versus 404 distinguishes existence. It is pre-existing and consistent with `OrganizationMemberAccess`, and exploiting it requires already holding a valid organization UUID, so the practical value is near zero. It is being tracked outside the issue tracker with a related authorization item, at the maintainer's preference. Happy to file it publicly if that is wanted — it is a hardening nit rather than a bypass.

Gates after all of the above: entity 48, entity_api 283, domain 238, web 158, 0 failures; clippy, fmt and `cargo check --workspace` clean. One clippy `needless_borrow` that my refactor introduced was caught and fixed rather than allowed.

jhodapp:
Correction to my previous comment: #3 is now filed publicly as #383, per the maintainer. It is scoped as hardening rather than a vulnerability, since it needs a valid organization UUID before the status code says anything.

The suggested fix there also removes a query on the denied path: check the role first, which reads the already-hydrated session roles and costs nothing, then resolve the organization only for callers who pass. Filed against both extractors so they keep the same shape.

jhodapp:
Good catch, and correct. Fixed in `8dfcc6ff`.

`goal.rs` resolved SSE recipients *after* the transaction committed and propagated failure with `?`. So a lookup failure returned an error for a mutation that had already succeeded: the caller saw a failure, no event fired, and the goal existed anyway. All three paths were affected — create, the shared update/update_status publisher, and delete.

The codebase had already settled this question elsewhere. `action.rs` carries the rule in a comment:

> Best-effort SSE notify set for an action [...] The DB write is the contract; a failed lookup must NOT fail the mutation, so log and return None.

`agreement.rs` does the same. `goal.rs` was the outlier, so the fix is to match them rather than invent anything: the helper returns `Option<Vec<Id>>`, logs on failure, and callers skip publishing. With no resolvable recipients there is nobody to notify, so skipping the event is the same outcome as publishing to an empty set.

Worth being precise about ownership: the `?` predates this PR, but membership filtering added a second query to that post-commit path, so this PR roughly doubled the failure surface of an existing fragility. Fair to call it in scope.

Covered by `create_succeeds_when_the_notify_lookup_fails`, which mocks a successful goal write followed by a failing relationship lookup and asserts both halves: the call still returns `Ok`, and no event is published. Mutation-checked by restoring the error propagation, which fails it.

domain 238 → 239, everything else unchanged, clippy/fmt/`cargo check --workspace` clean.

jhodapp:
Both addressed in `e582228b`, though #3 needed a different fix than proposed.

**#3 — the organization segment.** Two corrections to the premise, then a stronger fix.

First, this is not a regression from my change. The old signature was `Path((_organization_id, relationship_id))` — note the underscore. The segment was parsed and discarded, so it was decorative before too.

Second, and more important: adding `OrganizationMemberAccess` back would not have made it meaningful. That extractor checks the caller is a member of the *path* organization; it never compares the relationship's organization to it. So the sibling route in `actions_controller.rs` has the same gap — a caller who is a member of org A and a participant in a relationship in org B can still address it as `/organizations/A/coaching_relationships/{rel_in_B}/actions` and succeed. Its `OrganizationMemberAccess(_organization_id)` discards the value too. The two routes agreeing at the same weak level is not the same as the segment meaning something.

The fix is to compare `relationship.organization_id` against the path segment and 404 on mismatch. 404 rather than 403, so a wrong address does not confirm the relationship exists elsewhere. It costs **zero** extra queries: the relationship is already loaded and the segment is already parsed, versus the two queries `OrganizationMemberAccess` would have added.

Put it in the extractor rather than the one handler, so all three consuming routes get it — including the sibling that prompted the comparison. Covered by `read_denies_a_relationship_addressed_under_the_wrong_organization`, mutation-checked.

**#4 — the scan.** Done, and I was wrong to defer it. I had filed it as #382 because moving the filter into SQL disarms `does_not_reuse_a_reversed_pair`: `MockDatabase` does not apply a `WHERE`, so the test could no longer tell a matching pair from a reversed one.

The way out is to assert on the *bindings* rather than the rows. A SELECT renders with placeholders, so the SQL text alone cannot catch a value swap, but the mock's transaction log carries the bound values. The test now asserts organization, coach, coachee in that order, and swapping the two filter columns fails it. Verified by mutation. So the property survives the move, and the lookup is now a single indexed row.

Closes #382. entity_api 283, domain 239, web 161 → 162, entity 48, all green; clippy, fmt and `cargo check --workspace` clean.

jhodapp:
Confirmed and fixed in `6238e43b`.

Both unlink paths delete the link, then resolve recipients, then propagate with `?`. A membership-query failure returned an error for an unlink that had already committed and dropped the event, so the caller saw a failure and would retry against already-mutated state.

This one is more clearly mine than the goal-path version you flagged earlier. `publish_session_goal_deleted` **could not fail at all** before this PR — it built the recipient list inline with no query. Membership filtering added one, which is what put a fallible call after a committed write.

Rather than make it best-effort by convention, it is now **infallible by signature**: it returns `()`, so it cannot fail the mutation because it cannot fail. That is the property both callers actually need, and it holds without anyone having to remember it.

**Swept the rest of the class rather than wait for a third report.** Every site that resolves an SSE notify set:

| Site | Shape | Verdict |
|---|---|---|
| `goal.rs` create / update / delete | after commit, `?` | fixed in `8dfcc6ff` |
| `coaching_session_goal.rs` unlink x2 | after commit, `?` | fixed here |
| `coaching_session_goal.rs` link | **before** the transaction | safe by construction |
| `coaching_session_hydration.rs` x2 | inside `ctx.txn`, failure rolls the write back | safe |
| `find_participant_ids` consumers (12) | already `match`, best-effort | safe |

So the two you found were the only two.

The link path is worth calling out as the shape to copy: it resolves recipients *before* opening its transaction, so a lookup failure fails before anything is written. That is stronger than best-effort, and it is available wherever the recipients do not depend on the write.

Covered by `unlink_succeeds_when_the_notify_lookup_fails`, asserting both halves — the unlink still returns `Ok`, and no event fires. Mutation-checked by restoring the propagation.

domain 239 → 240, everything else unchanged, clippy/fmt/`cargo check --workspace` clean.

PR #385, opened 2026-08-11
Revoke the token with the provider on disconnect

jhodapp:
## Description

Disconnecting an OAuth integration deleted the `oauth_connections` row without telling the provider. Tokens stayed valid at the provider and the grant remained listed on the coach's account permissions page, so the disconnect was a lie from the user's point of view. `revoke_token()` was already implemented and had zero callers.

Blocks Google production verification for `meetings.space.created`, whose submitted scope justification states that tokens are revoked on disconnect. **That claim is now verified true** (see Testing Strategy).

Scope note: this fixes the Google path, which is verified end to end. It also corrects a Zoom bug found along the way, but Zoom revocation ships **unverified** because Zoom OAuth is not configured in any environment. Details below.

### Changes
* `domain::oauth_connection::delete_by_user_and_provider` is now a real domain function instead of a passthrough re-export. It revokes the stored grant, then deletes the row.
* Revocation is best effort. Missing encryption key, decrypt failure, unbuildable credentials, and a rejecting or unreachable endpoint are all logged and the row is deleted anyway. Leaving it behind would strand the user in a connected-but-broken state with no way to retry.
* Tries the refresh token, then the access token. Google accepts either; Zoom documents only the access token. Trying both keeps one code path correct for both providers.
* Resolves the connection once and deletes by id, closing a window where a concurrent disconnect returned 404 to the caller after the grant was already revoked.
* The already-revoked cleanup path in `get_valid_access_token` calls `entity_api` directly, so it makes no pointless revoke round trip.
* **Zoom revoke sent no client credentials at all**, where Zoom authenticates that endpoint with HTTP Basic. Unreachable until `revoke_token` gained a caller, so it had never been exercised. Fixed, but see the Zoom note under Testing Strategy before relying on it.
* No wire-format change. Revoke failures are deliberately not surfaced, so the client still sees success.

### Testing Strategy

Six domain tests. The load-bearing one asserts a `DELETE` actually runs when revocation cannot, verified by mutation; four cover token preference, fallback, and undecryptable input. Mock suite green (entity_api 283, domain 246, web 162); clippy and fmt clean.

No automated test reaches the wire, since providers are built from config rather than injected, so manual verification is the only proof revocation works. Plan at `docs/test-plans/oauth_disconnect_revokes_token_manual_testing.md`.

* **Google: PASS, verified 2026-08-11.** Connected Google, confirmed the app was listed at https://myaccount.google.com/permissions, disconnected, and the grant was gone from that page. This is the acceptance criterion behind the Google submission.
* **Zoom: BLOCKED, ships unverified.** `GET /oauth/zoom/authorize` returns 500 because `ZOOM_CLIENT_ID`, `ZOOM_CLIENT_SECRET` and `ZOOM_REDIRECT_URI` appear in **no** file in this repository, while their Google counterparts are in six. There is no Zoom connection to disconnect, so revocation cannot be exercised. Pre-existing and unrelated to this PR: Zoom OAuth has never been usable in any environment. Enabling it needs a Zoom Marketplace app and those three variables wired through every layer, tracked separately.

### Concerns
* A failed revoke is visible only as a `warn!`. Correct against blocking disconnect, but a broken revoke path fails silently and looks identical to success from the client, which is how the Zoom bug survived. Worth a log alert on `Failed to revoke`.
* The Zoom Basic auth fix follows Zoom's documented requirement but has never been observed working. It is strictly better than sending no credentials, and it should be re-tested when Zoom OAuth is configured.
* Anyone who disconnected before this ships still has a live grant and no row left to revoke from. They would have to revoke manually from their provider's account permissions page.

### DIFF
--- a/docs/test-plans/oauth_disconnect_revokes_token_manual_testing.md
+++ b/docs/test-plans/oauth_disconnect_revokes_token_manual_testing.md
@@ -40,19 +40,29 @@ account permissions page after disconnecting.
 Same shape as Scenario A, against Zoom's app management page
 (https://marketplace.zoom.us > Manage > Added Apps).
 
-This scenario carries the real risk. Zoom's revoke endpoint authenticates the
-client with HTTP Basic, and the request previously sent no client credentials at
-all, so it would have failed on every call. That code was unreachable until
-`revoke_token` gained a caller, so it had never run. The fix follows Zoom's
-documented requirement but has not been observed working.
+**Result: BLOCKED. Zoom OAuth is not configured in any environment.**
 
-Because failures are `warn!` only, a still-broken Zoom revoke looks identical to
-success in the UI. Check the backend log as well as the Zoom page.
+`GET /oauth/zoom/authorize` returns 500 because `create_zoom_provider` needs
+`ZOOM_CLIENT_ID`, `ZOOM_CLIENT_SECRET` and `ZOOM_REDIRECT_URI`, and none of the
+three appear anywhere in the repository: not in `.env`, `docker-compose.yaml`,
+`docker-compose.pr-preview.yaml`, `deploy_to_do.yml`,
+`ci-deploy-pr-preview.yml` or `docs/setup.md`. Their Google counterparts are in
+all six. There is no Zoom connection to disconnect, so revocation cannot be
+exercised.
+
+Enabling Zoom means creating a Zoom Marketplace app and wiring those three
+variables through every layer. Until then Zoom revocation ships **unverified**.
+
+When Zoom is enabled, run this scenario before trusting it. Zoom's revoke
+endpoint authenticates the client with HTTP Basic, and the request previously
+sent no client credentials at all, so it would have failed on every call. That
+code was unreachable until `revoke_token` gained a caller. The fix follows
+Zoom's documented requirement but has never been observed working, and because
+failures are `warn!` only, a still-broken revoke looks identical to success in
+the UI.
 
 **Pass:** the app is gone from Added Apps, and the log shows
 `Revoked zoom grant for user <id>` rather than `Failed to revoke`.
-
-**Result: NOT YET RUN.**
 
 ## 4. Scenario C: A failed revoke still disconnects the user
 
--- a/docs/test-plans/oauth_disconnect_revokes_token_manual_testing.md
+++ b/docs/test-plans/oauth_disconnect_revokes_token_manual_testing.md
@@ -1,0 +1,77 @@
+# Test Plan: Manually Testing OAuth Disconnect Revocation
+
+Verify that disconnecting a meeting integration revokes the grant with the
+provider, not just the row in our database.
+
+The single most important property: **the provider must forget us.** Before this
+change, Disconnect deleted the `oauth_connections` row and returned, leaving the
+access and refresh tokens valid and the grant listed on the coach's provider
+account page. Our app showing "not connected" is **not** evidence of revocation.
+The provider's own permissions page is the only evidence that counts.
+
+> [!IMPORTANT]
+> No automated test reaches the wire. Providers are built from config rather than
+> injected, so the suite covers token selection and the delete but never an actual
+> `revoke_token` call. These scenarios are the only proof revocation works.
+
+## 1. Prerequisites
+
+- Backend on branch `fix/oauth-disconnect-revokes-token`, migrations applied.
+- `ENCRYPTION_KEY` set, plus `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` /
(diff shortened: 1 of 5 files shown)


jhodapp:
Thanks. Fixed the P2 in `d0f1a9a`-ish (latest push); pushing back on the P1.

**P2 — deletion test never verifies deletion: valid, fixed.**

The test now inspects the mock statement log and asserts a `DELETE` ran. Confirmed it is not vacuous by mutation: commenting out the `delete_by_id` call makes it fail with the intended message.

The comment is corrected too, though not quite as described. `Config::default()` is `from_args([...])` and every field is `#[arg(long, env)]`, so whether an encryption key is present depends on the ambient environment rather than being absent by definition. Both branches bail before any network call, which is the property the test actually relies on, so the comment now says that instead of naming one branch.

**P1 — stale disconnect deletes refreshed credentials: real race, but not introduced here, and the obvious fix is worse.**

Agreed the window exists: we resolve the row, await the revoke, then delete by the captured id. Two things though.

*It is not a regression.* The previous code called `delete_by_user_and_provider`, which re-resolves by `(user_id, provider)` and deletes whatever it finds. A concurrent refresh (`update_tokens`, `entity_api/src/oauth_connection.rs:85`) and a concurrent reconnect (`exchange_and_store_tokens` takes the `Some(conn)` branch and calls `update_tokens(db, conn.id, ...)`) both write to the **same row id**. So the old by-pair delete removed exactly the same row the new by-id delete removes. Delete-by-id is strictly safer in the one case where they diverge: if the row were deleted and recreated with a fresh id, the by-pair lookup would delete the new row while the by-id delete correctly hits nothing.

*The fix would trade a rare bad outcome for a common one.* Guarding the delete on an unchanged `updated_at` means a background token refresh landing mid-revoke aborts the disconnect. The user clicked Disconnect, we already revoked the grant, and they would be left with a live row pointing at a dead grant, which is the connected-but-broken state the doc comment exists to prevent. Background refresh is far more likely than a deliberate reconnect during a disconnect's network window.

Worth noting the refresh case is not actually harmful: revocation kills the whole grant at the provider, so tokens minted moments earlier are already dead and deleting the row is the correct end state. Only the reconnect-during-disconnect case loses anything, it requires re-consenting inside a sub-second window, and it self-corrects by reconnecting again.

Leaving as is deliberately. Happy to revisit if you see a variant that does not sacrifice the common path.

jhodapp:
Re the 4/5 and "a concurrent refresh or reconnect can be deleted after it succeeds" — I traced this properly rather than re-asserting my earlier reply, and the conclusion is stronger than last time: **delete-by-id is not the cause, and it is strictly safer than the code it replaced.**

**Every writer to an existing row keeps the same id.** There are exactly two paths that can write to a live connection while revocation is in flight:

- refresh, `domain/src/oauth_token_storage.rs:75` -> `update_tokens(self.db, conn.id, ...)`
- reconnect, `domain/src/oauth_connection.rs:99` -> `update_tokens(db, conn.id, ...)`

Both take the `Some(conn)` branch, and `update_tokens` sets `id: Unchanged(existing.id)` (`entity_api/src/oauth_connection.rs:100`). The row id never moves.

So the previous code, which re-resolved by `(user_id, provider)` and deleted whatever it found, deleted **exactly the same row** that delete-by-id deletes. The captured id is not what makes the write vulnerable; the write lands on the row either way. This race predates the PR unchanged.

**Where the two do diverge, delete-by-id is the safe one.** `oauth_token_storage.rs:85` has a `None` branch that creates a row with a **fresh id**. That makes this reachable:

1. Disconnect A loads row X and starts revoking.
2. Disconnect B completes and deletes X.
3. A refresh runs, finds nothing, creates row Y with fresh credentials.
4. Disconnect A finishes.

Old code deletes by pair: finds Y, destroys the newly created credentials. New code deletes by id X, which is gone, and does nothing. This PR fixes that case rather than causing it.

**On the residual race, deleting is the correct outcome, not a bug.** Revocation kills the whole grant at the provider, so tokens written moments earlier are already dead. Keeping the row would leave a live-looking connection backed by a revoked grant, which is the connected-but-broken state the function exists to avoid. Only reconnect-during-disconnect loses anything, it requires re-consenting inside the revoke's network window, and it self-corrects by reconnecting again.

**The obvious guard makes things worse.** Conditioning the delete on an unchanged `updated_at` means a routine background refresh aborts the disconnect: the user clicked Disconnect, we already revoked the grant, and they keep a row pointing at a dead one until something happens to notice. That converts a rare, visible, self-service failure into a more common, invisible one, and background refresh is far likelier than a reconnect race.

I have added a line to the doc comment so the behavior is stated rather than implied (`0b1e0b6`-ish, latest push). Treating this as resolved. If you have a variant that closes the window without sacrificing the disconnect on a routine refresh, I would genuinely like to see it.

